PR #241, opened 2023-01-18
Trust-kpis

eatyourpeas:
This implements a table on selected hospital page of nhs trust kpis next to national KPIs
The national KPIs are intensive on the database and will need reworking for production. This first attempt is a thought experiment only at this stage


pacharanero:
Nice work @eatyourpeas

Issue #159, opened 2022-10-18
Design feedback on E12 platform from RCPCH design team


pacharanero:
|Recommended change|Type|Comments|
|---|---|---|
| Change all font to Montserrat, including for copy text, menu items|Design|Some has changed; could also change buttons like 'View all XYZ children' on clinian dashboard, which is now Helvetical. Presume want to use Arial for data, which is fine|
|Mobile display - Menu does display, but optionally use a burger menu to hide it|Design|Simon confirmed unlikely to be heavily used on mobile, but can work to improve mobile display esp on front page, other content pages|
|Colours - White text on our bright blue (#11a7f2) does not pass accessibility colour contrast – needs to be black text. Or use white background with Dark blue text for large blocks of copy|Design|Complete|
|Use sentence case for menu items, so ‘Patient access’ not 'Patient Access', etc.|Editorial|Complete|
|Don't use full stop after headings, eg 'The UK National Audit for Childhood Epilepsies'|Editorial|Complete|
|Home page content - To write new, welcoming content about who and what the site is for, and what you can do here? Can add a link to other pages for more detail as needed|Editorial|C&B/Ep12 team to provide mockup with copy / C&B team happy to review|
|Branding - There are two Epilepsy12 logos - new one at top banner, and old one in content window. Full RCPCH logo should be used. Recommend removing Epilepsy12 and RCPCH acronmyn logos from top of content window, even removing that top content box. And add logos - including full version to a footer, if available, and with RCPCH logo linking to main website|Layout/navigation|C&B team and Epilepsy12 team discussing options for revised Epilepsy12 branding, with one logo version possibly having strapline - will provide and can test that displays well alongside RCPCH logo|
|Links for clinician access and patient access - move higher up home page|Layout/navigation|Complete|
|Community & hospital' - Use 'and' instead of ampersand, and to match 'Children and families' - Generally we recommend using 'and' but not a major thing!|Editorial||
|Menu - Add Home link in addition to RCPCH/E12 logo/text?|Layout/navigation|A convention we use on our other sites, but not essential|
|Menu / layout - This cuts off for Kirsten's view, but possibly only as I see 'Database management' which I assume most users won't see. See screenshot Check to see if issue for others; if so, reduce font size?|Design||
[...]

PR #248, opened 2023-02-01
Adds Favicon to the site

pacharanero:
A small addition but for some reason annoyingly difficult to get working for various reasons.

Can be seen in action at https://epilepsy12-development.azurewebsites.net/ (which is otherwise working but has no data in the DB yet so won't be able to log in)

Progress towards #159 (which is a large issue that could probably do with a full check-through and then Design Team could review the E12 app again)

This PR also tidies up the GH workflows in the Development branch. It should be a separate PR really.


eatyourpeas (approved):
That looks good to me

PR #270, opened 2023-03-09
Overhauled the footer styling

anchit-chandran:
- moved all obvious footer stylings into `footer.css`
- refactored semanti UI structure of `footer.html` into more `container>row>col` structure


eatyourpeas (approved):
Thanks Anchit - this looks fine.

PR #284, opened 2023-03-11
LSOA postcode error

### LINKED ISSUE #185: LSOA postcode error
If the postcode is not valid there is currently no error message raised 
it does not break the object saving but the LSOA look up and IMD calculation then fails.
- [ ] Need to validate postcode using api.postcodes.io prior to save

eatyourpeas:
Implements postcodes for unknown address or no fixed about using standard NHS codes
Add toggle button to create_case and update_case form and templates
hyperscript for hide/show
Fixes #185


eatyourpeas:
@pacharanero I would be grateful if you could review this when you get the chance.
1. update hyperscript package 
2. add new htmx templates with hypscript to allow toggle buttons for unknown postcodes in the create_case template
3. add new logic in the cases_view.py to accommodate this 
4. fix the failure to redirect to cases form on save or update case
5. fix the filter box in case list

pacharanero:
OK will take a look - this week I'm at a conference but will see what I can do

pacharanero (approved):

PR #307, opened 2023-03-22
Eatyourpeas/issue302

eatyourpeas:
adds inline styles to emails with base64 conversion of RCPCH logo
pulls in montserrat from googlefonts


eatyourpeas:
This can only really be tested when DEBUG=False, so merging to development for this purpose.

PR #485, opened 2023-05-09
fixes a minor typo in the antiepilepsy_medicine.py model, and also fixes it in the initial migration to save having messy migrations.

pacharanero:


pacharanero:
Should be a simple merge

eatyourpeas (approved):

Issue #493, opened 2023-05-22
Epilepsy type select boxes broken


anchit-chandran:
When selecting `Epileptic` for seizure type, it automatically also selects `Non-epileptic`, and doesn't allow you to unselect either.

It works when selecting any of the other options.

![Image](https://user-images.githubusercontent.com/68898006/239877387-258a585b-3bc7-4697-b1fa-ec347da1a23b.png)

eatyourpeas:
fixed by #494

PR #494, opened 2023-05-22
Validators

eatyourpeas:


eatyourpeas:
fixes #493 
adds validators.py for dates

PR #510, opened 2023-06-14
Finished KPI Tests

anchit-chandran:
This PR finishes work on KPI tests, which now all score properly. KPI score functions were updated if any odd behaviours were found.

It also refactors the Factories, so that answers can automatically be set from the Case level, with determined KPI scores.

NOTE: the PR also includes a tiny bit of the start of perms testing, but nothing significant


pacharanero (approved):
Pulled branch, rebuilt Docker, ran tests and all expected passing tests did pass (not including xfails etc). Happy to merge to dev

PR #517, opened 2023-07-07
*based on PR #510 'Finished KPI Tests'* Finishes first round of testing

anchit-chandran:
Additional test areas from previous PR include the following points, alongside refactors where tests were implemented to ensure regressions are prevented.

# 1 Permissions testing

- For each user role, specified by the E12 team, we assert they can / can't access various url endpoints, receiving the appropriate status code

# 2 KPI Aggregation

- For the main 'Organisation' page which shows summary statistics for KPIs across the E12 audit, we have implemented tests to assert the aggregations output the expected numbers, using seeded Cases with known KPI pass/fail/ineligible outcomes. Particuarly for the entire KPI aggregation function, we use 100 Cases, with all tests passing.

# 3 Form calculation 

- We have implemented tests to ensure, on a model basis, that both numerator and denominators denoting audit progress output the correct values, given known inputs. This includes cases where certain answers cause different numbers of options to be available, such as with new `Episodes`


pacharanero (approved):
Tested locally & all passing

PR #554, opened 2023-08-29
Add PtestSuite to run on PRs

anchit-chandran:
Similar workflow to autodeploys, but runs on PR instead of after merge.

This helps promote safe, high-quality software dev practices + minimise human errors. Bypassing the pytests / reviewer checks CAN be done in a "break glass for emergency" situation, but will need to be a conscious decision, against Incubator 'norms'

Also changes git hash solution to combo of mine and @pacharanero :
- In local development, will look inside .git folder
- In environments where no .git folder exists, e.g. the GitHubActions environments for build + deploys, those workflows add a git hash .txt file, from which the git info can be read

**NOTE**: updated the git hash to also be visible to rcpch staff / audit members. found it would be helpful in cases where they report a bug that *should* be fixed, but unsure what version they're looking at.


anchit-chandran:
Tests currently failing because:

1. Census platform returning None -> is the secrets env var updated?
2. Git branch context processor can't find the .git/ref file @pacharanero unsure how your new solution for that works - might need your help coming up with error handling for running this inside GHActions tests

eatyourpeas (approved):

PR #565, opened 2023-09-07
Significant refactor to include geographical boundaries

anchit-chandran:
**NOTE**: ❗this is *expected* to break development due to the significant model refactors, however, with the move to a VPS, this update will be necessary anyway

Significant refactor to all geographical entity models to include more details.

Took this opportunity to also homogenise naming of multiple model attributes


eatyourpeas (approved):
This is a large PR and has fixes for several other issues in it besides the refactor. 
Main thing is all tests are passing.

Issue #424, opened 2023-04-25
fetch_snomed.py can be refactored to remove duplication


pacharanero:
Numerous similar functions in this file.

1. Many are unused and can simply be removed.

2. If there are any duplicated lines of code in what remains then a refactor may be possible to reuse one snomed lookup function

anchit-chandran:
closing as fixed in #569

PR #569, opened 2023-09-09
prune general functions

eatyourpeas:
implement nhs_number package
prunes all unused functions from general functions folder


eatyourpeas:
incidentally, merging this would also close #424

anchit-chandran (approved):

PR #574, opened 2023-09-15
adds captcha, small 2fa improvements

anchit-chandran:
Adds captcha to login page.

NOTE: if settings.DEBUG == True, then captcha value is automatically inserted into field (so can just quickly click through) (if settings.CAPTCHA_TEST_MODE == True, then "PASSED" is accepted)

Also:

- extended the django-2fa LoginView to redirect to org summary page if successful.
- removed previously only commented out code for prev login view

image


eatyourpeas (approved):
This is great - thanks Anchit

PR #651, opened 2023-11-23
Cohort-data

### LINKED ISSUE #641: Cohort calculation error?
Can I check that the cohort calculation is definitely correct? Sorry I know we have previously covered it:
image

Here, both children are allocated to cohort 6, but the first has 51 days left from a first paediatric assessment date of 15/12/22,
while the second has 422 days left with a first paediatric assessment date on 17/7/2023. Is it ok they are both still in the same cohort?

eatyourpeas:
refactor cohort calculation to seeding

### Overview

This PR fixes 2 issues with the cohort calculations
1. the cohort data submission date was wrong (out by a year in some cases)
2. the one year completed date was wrong

### Code changes

rewriting of the cohort calculations with 2 functions
1. cohort_number_from_first_paediatric_assessment_date
2. dates_for_cohort

All functions using the previous get_current_cohort were refactored

The registration_close_date was refactored to completed_first_year_of_care_date and the in the save() method for Registration

Updates were made to the UI where completed_first_year_of_care_date was referenced

completed_first_year_of_care_date was added to fields to discount when scoring the form

Tests were also wrong in some cases and these have been corrected. Currently all passing

Please describe any changes to documentation here.

As this is a bug fix, it does not really have any implications for documentation

### Related Issues

closes #641
Note that new migrations have been made in this fix

### Mentions

@pacharanero


anchit-chandran (approved):
LGTM, tests passing

PR #667, opened 2023-11-29
643 end point for e12 staff to download patient data

### LINKED ISSUE #643: End Point for E12 staff to download patient data
As agreed in #534 the E12 team need a way to download the data from patient records to allow our annual analyses and reporting. This can either be via a download button to give a large csv with one patient per row, or via the API. The access to this will be restricted to E12 staff only.

The data can be pseudonymised data, where the following personal identifiers can be removed/modified before download:
- first and last name, NHS number removed
- DOB transformed to age at first assessment; in months for those aged 0-2yrs, in years for those aged over 2 years
- Home postcode transformed to LSOA deprivation quintile

The above is also specified in the DPIA #625 

This will be required in March 2024 (post-launch).

dc2007git:
Creates a feature to address the issue in #643.

A .csv file is created which contains 4 columns - sex, deprivation quintile (calculated from case's postcode), date of first assessment and age of case at first assessment.

Also adds button to bottom of case cohort list to download this csv, where permitted users are the rcpch audit team.

Please note that further testing is yet to be implemented which will provide insurance for intended use.

Update - I've just seen @nikyraja 's comment on #643 , and pending a meeting will add more features and columns as requested from the E12 team.

Button:
image

.csv contents:
image

### DIFF
--- a/epilepsy12/general_functions/__init__.py
+++ b/epilepsy12/general_functions/__init__.py
@@ -5,7 +5,7 @@ from .cohort_number import *
 from .index_multiple_deprivation import *
 from .postcode import *
 from .fuzzy_matching import *
-from .time_elapsed import stringify_time_elapsed
+from .time_elapsed import stringify_time_elapsed, rcpch_requested_time_elapsed
 from .value_from_key import *
 from .date_functions import *
 from .construct_confirm_email import *
--- a/epilepsy12/general_functions/time_elapsed.py
+++ b/epilepsy12/general_functions/time_elapsed.py
@@ -45,3 +45,36 @@ def stringify_time_elapsed(start_date, end_date):
 
     else:
         raise ValueError("Both start and end dates must be provided")
+
+def rcpch_requested_time_elapsed(start_date, end_date):
+    """
+    Calculated field. Returns time elapsed between two dates.
+
+    If less than 2 years of difference, time elapsed is returned in months
+    
+    If greater than 2 years of difference, time elapseed is returned in years
+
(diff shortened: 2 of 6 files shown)


eatyourpeas:
thanks for this @dc2007git 
I have not yet fully tested this but a couple of items strike me before I do this.
1. ```first_assessment_age``` column - my guess is that the E12 team want this metric as an integer or a decimal, rather than a string? Perhaps @nikyraja can confirm?
2. Rather than test for the group at the top of the function, we would do better to do this in a decorator to sit below the ```@login_and_otp_required()``` decorator. There is already a deprecated ```has_group``` decorator in the code base that  I think is nolonger used - so you could use that. But my suggestion, which is better design but more work, would be to create a new custom permission 'can_download_patient_data' in the Epilepsy12User model meta class. If you go this route, you would need to update the ```groups_seeder``` function so that the permission is added to the E12 full access group by default on seeding in future, for example when working locally or in dev or staging. But actually, live is already seeded now, so we would have to makemigrations and migrate, and then in django admin manually add the new permission to the E12 audit team group for it to be active in live. You could then test ```perms.epilepsy12.can_download_patient_data``` in the template to conditionally show the button.

eatyourpeas:
and by the way....
If you are calculating years as a decimal, our practice in the growth charts is to divide the difference between two dates in days by 365.25 to get decimal years. The extra 0.25 is because every 4th year is a leap year.

mbarton:
@dc2007git I'll close this PR as it's been open for a while. Feel free to re-open if it's still relevant but I think we'll probably look at search by case id https://github.com/rcpch/rcpch-audit-engine/issues/799#issuecomment-1961545693 first

PR #676, opened 2023-12-01
Registration fix, KPIAgg model Test Update, Cohort days remaining until last submission UI Bug Fix 

anchit-chandran:
### Overview

- **Critical change @eatyourpeas**: `Registration.first_paediatric_assessment_date` is required to set a Case's Cohort, and to validate the inputted FPA date for a newly registeredCase. This won't be set before a Case is first registered. Therefore, Cases currently cannot be registered. The fix involved calculating this date not from `Registration.first_paediatric_assessment_date` (which will be None), but instead from the inputted form field value for the FPA date. 
- Test update: KPI Aggregation models are now stratified based on cohort, so we cannot simply get a model based on an abstraction code, we must also specify the cohort
- Fix the cohort days remaining until latest submission not appearing

### Code changes

- **Critical change** see above
- Test updates
- Simply update template HTML to reference correct attribute

-

### Related Issues

#676 - UI bug fix (see screenshot)
image

### DIFF
--- a/epilepsy12/common_view_functions/validate_form_update_model.py
+++ b/epilepsy12/common_view_functions/validate_form_update_model.py
@@ -12,6 +12,7 @@ from psycopg2 import DatabaseError
 # RCPCH imports
 from ..general_functions import (
     dates_for_cohort,
+    cohort_number_from_first_paediatric_assessment_date,
 )
 from ..validators import epilepsy12_date_validator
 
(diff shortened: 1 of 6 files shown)


pacharanero:
Thanks for such a fast fix @anchit-chandran.
This is part of the application I don't have much familiarity with, so would appreciate @eatyourpeas to review. I'm happy to join a call in which we go through the changes, but I won't be able to do this over the weekend.

eatyourpeas:
Thank you @anchit-chandran for pinning this down. So the cohort now is set in the model in the on save method, but of course, as you identified, the cohort, and therefore the cohort end date is necessary for the error handling before this step to make sure the date supplied is not after the cohort end date or before the cohort start date. So the fix you applied was correct.

Incidentally I discovered ironically by chance a bug in the random_date function that would normally generate a random date between today (current date) and beginning of current cohort (which is always the first of December). An by 1/365 chance, today is the 1st December and so the only day in the year that running this function would fail.

Anyway it is now fixed, so merging and closing this now.

PR #711, opened 2023-12-09
Eatyourpeas/issue710

### LINKED ISSUE #170: Dummy data - do we need a better source of seeding data for a testing phase?

### LINKED ISSUE #710: KPI individual measures navigation buttons viewable even when no data
When there is no data in the individual measures segment, the navigation buttons which navigate between the graphs should not be visible.

image

eatyourpeas:
organisation view now hides tables and shows advisory if no data

### Overview

The Organisation View currently a little messy if the organisation selected in the view has registered no patients. This PR hides the table that shows the KPIs across levels of abstraction, as well as the table which looks at individual KPIs.
Instead an advisory message is shown saying there is no data to display.

As part of this PR it became clear that in development it is helpful to seed specific cohorts, rather than always seed just the currently recruiting cohort. An additional attribute has been added to manage.py seed --mode=seed_registrations --cohort=6

Finally, an earlier recent change in the token template led to an error which resolved on removing the autocomplete off template tag

### Code changes

Changes are made to selected_organisation_summary.html with a conditional to test if data present to render the tables, and if not to show a new advisory ui message

in kpis.html, individual_metrics.html and metrics.html the previous conditions that did this imperfectly were removed

seed.py has the new attributes and params for seeding by cohort number

organisation_views.py has had some comments removed

this is largely stylistic. I will update the documentation regarding seeding as separate commit after this

### Related Issues

Closes #710


eatyourpeas:
image

pacharanero:
edited as the Issue linked was #170 instead of #710 which would have closed the wrong issue on merging

pacharanero (commented):
Reviewed and this PR fixes #710

Issue #538, opened 2023-08-03
Patient transfer notification emails


AmaniKrayemRCPCH:
I tested out a patient transfer to see if the permissions were correct for clinician users (see [issue #536](https://github.com/rcpch/rcpch-audit-engine/issues/536)) and a notification email appeared to have been sent to everyone, including staff not allocated to the site. 

The email also contains the patient's name. We'd much prefer an anonymous ID of some sort, as we can't guarantee that the recipient's email is secure to handle patient identifiable info. Plus, if RCPCH staff receive the email, we can't handle patient info in our servers. 

![Image](https://user-images.githubusercontent.com/126471383/258144947-78b24d26-6f84-49dc-8adf-bf4e18e1df24.PNG)

AmaniKrayemRCPCH:
email was also sent to nnap@rcpch.ac.uk, and this account wasn't 'active' at the time.

eatyourpeas:
Thanks for testing this functionality.
We have not fully discussed the workflow of this yet, so I was not clear how you wanted this to work.
Currently emails are sent to the lead clinician of the accepting trust if they exist and all E12 team. Perhaps, rather than an identifier, it should simply say 'Dr X from Organisation Y has transferred a young person to your team. Please contact Dr X for further details...'

Do please advise on appropriate text and recipients.

pacharanero:
* [x] transfer buttin is going to be removed see #513 
* [x] remove patient's name from the transfer email anyway
* [x] transfer emails should go to E12 team email address epilepsy12@rcpch.ac.uk

AmaniKrayemRCPCH:
Has this been actioned?

eatyourpeas:
Yes this has been actioned

nikyraja:
A patient was transferred between organisations and the project team were included in the email. 

Only the clinicians should receive these transfer emails. If there is no clinician registered to the receiving organisation, the email can go to [epilepsy12@rcpch.ac.uk](mailto:epilepsy12@rcpch.ac.uk). No transfer emails need to be going to personal emails. 

@eatyourpeas @pacharanero

eatyourpeas:
Apologies - in the last fix all RCPCH audit team members were included in the transfer request emails. These have been removed in pull request #726 which should close this issue.
In addition I have made some small extra changes

nikyraja:
Great, thank you Simon!

PR #726, opened 2023-12-20
Eatyourpeas/issue538

eatyourpeas:
### Overview

This fixes 4 things to do with the workflow for transferring children between centres.
1. remove E12 users from the email cc when the transfer request is sent
2. if the transfer is refused and the child is returned to the original trust, fix the bug that the original sending organisation is added to list of historical trusts.
3. change the reject/accept modal to read 'I confirm' rather than the more ambiguous 'I accept'
4. Remove the current organisation from the dropdown which appears when the transfer button is pressed. This prevents the user from transferring a child into the organisation where they already are.

### Code changes

Changes are made in 3 files
case_views.py - this deletes the new site object created if the transfer is rejected, and reanimates the previous record.

registration_views: removes members of the RCPCH audit team from the cc in tranfer requests email. removes current organisation from transfer organisation list

case_table: change the wording in the modal button

Bug fix, so no implications for documentation

### Related Issues

#538

### Mentions

@anchit_chandran @pacharanero

### DIFF
--- a/epilepsy12/views/assessment_views.py
+++ b/epilepsy12/views/assessment_views.py
@@ -9,6 +9,135 @@ from ..common_view_functions import (
 from ..decorator import user_may_view_this_child, login_and_otp_required
 
 
+def update_site_model(
+    centre_role: str, selected_organisation, case, user, site_id=None
+):
+    """
+    Helper function to update sites model with attributes describing role of site
+    (general paediatric, neurology, surgical centre)
+    This is complicated because historical records of lead centre status are persisted
+    whereas historical records of other roles are not.
+
+    If the organisation is the lead site for this child, and also the neurology/surgery
+    or general paediatric centre, all this information can be stored in one record.
+
+    If the organisation used to be the lead site for this child, and now is actively
+    the neurology/surgery or general paediatric centre, two records are stored: one where
+    site_is_actively_involved_in_epilepsy_care is False and site_is_primary_centre_of_epilepsy_care is True,
+    the other where site_is_actively_involved_in_epilepsy_care is True and one/some of the other attributes
+    is True. site_is_primary_centre_of_epilepsy_care here is False.
+    """
+
+    if centre_role == "general_paediatric_centre":
+        update_field = {"site_is_general_paediatric_centre": True}
+        update_fields = {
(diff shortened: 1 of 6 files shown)


eatyourpeas:
Forgive me for piggybacking these changes on the back of this PR. It is related and equally urgent, so thought it might acceptable to couple the two issues.

In addition to the items/fixes listed in the PR above, subsequent commits have:
1. removed a filter bug where local health boards did not see their organisations' users
2. removed the empty brackets in the footer segment in case form

These items are in `templates/epilepsy12/forms/case.html` and `views/user_management_views.py`

The main fix is a bigger issue though:

## The problem
Currently if a child is transferred between centres, if the transfer is rejected (or sometimes accepted), the relationships it has with other centres in the Assessment tab which track general paediatric, neurology and surgical care are lost.

### Background
The Site model, as the middle table in the M2M relationship between Case and Organisation, stores information about the relationship a child has with an organisation. For each organisation involved in a child's care, either current or historic, a record is stored in the Site model. Flags document if a Site is actively involved in care, and the nature of the relationship (lead site, general paediatric/neuro or surgical site). To build a chronology of sites that have historically led the care of a child's epilepsy, it is possible to have multiple records for different organisations where lead site status is true, but active involvement status is false. These appear in the registration form as historical lead sites if they exist.

By contrast, any other relationships the site may have (not lead status, but provide general paediatric, neurology or epilepsy surgery care) are stored only once - no chronology for these exist. If a child moves its surgical care, a new record is created reflecting the change and the old one deleted.

It becomes more complicated if:
1. A lead centre transfers lead centre status to another organisation which is already involved in the care of the child
2. A lead centre transfers lead centre status to another organisation but continues to offer general paediatric, neurology or surgical care
3. Any requests for transfer are then rejected - as everything needs to be put back the way it was before the request was made

Having spent some time looking at these edge cases I am fairly sure this now works the way it was intended but really we need some tests. I have not had time so far to add these, and in the interests of urgency, am submitting this fix , with a view to adding some tests later on. I am really happy to help with the review of this PR as it is larger and more complicated than I would like.

pacharanero (commented):
Reviewed with @dc2007git and @eatyourpeas - tested the changes to lead site transfer logic with success.
Closes #538

PR #733, opened 2024-01-08
Upgrades to Live from Staging - for E12 Team Review

pacharanero:
### Overview

We have fixed some of the recent issues, these have been merged into development and staging. As per plan, discussed last week at Sprint Planning Meeting, we will await approval from E12 team that they are happy and have tested the workflows and verified the issues are fixed, before we merge into Live.

No need to review the code - just log into Staging and check the features are fixed in the UI.

Once E12 team are happy they can comment on this PR to signify approval and we will merge into Live. @nikyraja @AmaniKrayemRCPCH 

* NHS number visibility to local administrators has been fixed
* Transfer emails going to all E12 team has been fixed
* Logic of transferring site has been amended so that one cannot transfer a case to the same site.


nikyraja:
Thanks all for making the changes and Marcus as well for initiating this work flow for changes. 

We have had a look at staging and are happy that the above fixes have been made. Please can these be pushed to the live site. 

Thanks again.

pacharanero:
E12 live has been updated and checked as working

PR #740, opened 2024-01-10
filter and search case table in django admin

eatyourpeas:
### Overview

Django admin patients are neither alphabetical order nor searchable
This has been added

### Code changes

admin.py includes a new admin.ModelAdmin class for Case.
Adding this does lose the SimpleHistory admin though for this class as it is not possible to have both.

No implications

### Related Issues
No related issues

### Mentions

@pacharanero


eatyourpeas (commented):
Adding this because we have had some recent issues where Children have been removed through bugs from their lead organisation and we have had to search for them in the shell. This is a stop gap solution. Note that it includes a new migration as it adds ordering to the Case model

eatyourpeas:
I have added an extra commit to this PR - it sets the ODS code in the Organisation model to enforce uniqueness. Going forward if we are adding new organisations, we need to be careful not to add duplicates. I have tried this (given the problems with adding new fields) locally where there is an existing database and it does not break anything.

eatyourpeas:
I have fixed #745 which is a very small error but quite important in these commits.
The rest of the commits are to tidy up the admin - so that items are in order and reference the patient they are associated with. There were some areas where the detail view was broken because filtering against health board rather than trust was not made explicit. E12 team have needed more access to admin than I had expected so this is to make their lives easier until we can meet the issues raised in better ways.

**Note**
In the process I have by mistake fixed the logging string errors in the migrations which @pacharanero previously fixed (this is an older branch) and this has caused a conflict - all of these commits relating to conflicting files can all be safely ignored and if so then there should be no further problems.

pacharanero:
### Admin UI
@eatyourpeas sorry but after having trial-merged this in order to fix the merge conflicts, we noticed that (due to the older base of the `admin-improvements` branch) we had conflicting/duplicated migrations **0022** and **0023**. This meant we would have ended up with problems with migrations when this ran in dev/staging/live. The Admin UI changes are going to need to have `development` merged into them, merge conflicts fixed, delete the duplicate migrations and and then `makemigrations` again. Perhaps we will have to review and merge this in a call sometime.

### Syndrome picker problem #745 
We then tried cherry-picking the specific commit #22e631a2af5b5dbf5489a217736308d391476933 which fixes the Syndrome picker, because this is a fairly urgent fix.

However when we did this we still couldn't save the Syndrome, because for some reason the Date field which is associated with it gets overwritten before there is chance to Save. I tried reverting the recent change to the htmx delay time to 100ms but this still _didn't_ result in the date being saved. From experimenting on Live it looks like dates **are** validated and saved on Live.

Looking at commits between now and then I cannot work out what has changed, but it seems that the HTMX validation call fails for date and resets it. It's very unclear why in Live, the date field does save and the syndrome doesn't, but in the current commit for the Syndrome fix it is the other way around, the date field doesn't save (maybe fails validation?) and the syndrome name now DOES save.

I've recorded a litle screen recording below just to show it really did happen.
https://github.com/rcpch/rcpch-audit-engine/assets/2348385/52926296-7d72-48a0-9c76-906cf8fc6c1e

PR #758, opened 2024-01-19
Extend timeout in debug mode to reduce re logging in

pacharanero:
### Overview

1. to reduce the amount of time spent logging in when in local development by moving this setting into an env var
2. to reduce the complexity of adding new env vars to GitHub by making them all handled in a single GitHub secret and piped into a file for docker compose when the test runner starts.

### DIFF
--- a/.github/workflows/run-docker-compose-test-on-pr.yml
+++ b/.github/workflows/run-docker-compose-test-on-pr.yml
@@ -25,10 +25,6 @@ jobs:
           echo -e "${{ secrets.ENVIRONMENT }}" | base64 -d > envs/.env
           cat envs/.env
 
-      - name: Creates git_hash.txt file current commit hash
-        id: hash
-        run: echo "GIT_BRANCH=$(echo ${GITHUB_REF#refs/heads/}),GIT_HASH=$(git rev-parse --short "$GITHUB_SHA")" > git_hash.txt
-
       - name: Install Docker Compose
         run: |
           sudo apt-get update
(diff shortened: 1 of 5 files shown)


pacharanero:
Anyone who fancies reviewing this small PR is welcome to @anchit-chandran @eatyourpeas  @dc2007git 

It started off as simply adding an env var so that we can set logout timing according to the environment. In my local dev env I have set it to 36,000 minutes (25 days), because I hate endless captchæ and 2FA.

But then I decided to try for a stretch goal of getting rid of the many individual env vars that have to be added to GitHub secrets.

Eventually I found that it could be done if we base64 encode the data in a GitHub secret, and decode it when it's read out into the .env file in the GitHub action.

So, now we can get rid of all the myriad secrets and just have one. I am going to add a script that allows us to easily sync the local envs/.env file with the secret using `gh` the cli client for GitHub.

To do:
* delete all the other secrets except ENVIRONMENT
* add the AUTO_LOG_OUT_DELAY_MINUTES env var to dev, staging and live VPSs
* Implement `gh` script

eatyourpeas:
This is ingenious @pacharanero, lovely idea. I am finding that it does not seem to prevent the need for entering a token in development though, or am I using it wrong. Also a single test fails on github, but not locally? It focuses on `'active_git_branch'` being referenced before set, probably in `git_context_processor.py`. Perhaps we could get this over the line together?

pacharanero:
Yeah there's probably that one test to fix, then it'll be ready.

You will still need to enter a token in local dev, just you won't need to do it again routinely for as long as you set in .env (unless you reset the server and clear sessions)

pacharanero:
I'm going to merge this since it significantly improves the way env vars are handled in GH Actions.

IMPORTANT:
A new env var is required in all environments:

`AUTO_LOG_OUT_DELAY_MINUTES=30`
(all production environments will be set to 30 minutes, but in local dev you can set anything you prefer for convenience)

PR #763, opened 2024-01-23
Fixes the redirect to Signin Page problem

### LINKED ISSUE #748: Redirect to Sign In page when a user is logged out?
Just noticed that when one is logged out after the inactivity time expires, the page one is shown has no function:

image

It would be better to just send users back to the Sign In page, which will save them a click, and I think users understand when they see a Sign In page that they must have been logged out automatically, this happens with banking apps and other secure portals.

image

dc2007git:
### Code changes

Instead of redirecting the user to the 'additional privileges' template upon a 403 being raised, it redirects the user straight to the Sign In page. This has been tested by changing the AUTO_LOGOUT IDLE_TIME parameter to 1 minute and watching the Sign In page load in. Additionally, the 'You have been automatically logged out as there was no activity for 30 minutes' message pops up when this happens.

Now if the user is authenticated, then the 403 template is rendered, which has inbuilt conditionals depending on whether the user has been 2fa-verified or not, and deals with those situations. 

OR, if the user is not authenticated, and tries to access something they don't have permissions for, say a different organisation, then they face the 'additional privileges' page. Alternatively, upon session timeout, they are redirected to the login page

None

### Related Issues

Fixes #748

### DIFF
--- a/epilepsy12/views/redirects.py
+++ b/epilepsy12/views/redirects.py
@@ -22,9 +22,21 @@ def rcpch_403(request, exception):
     # the 403 template would be inserted into the target. This way the HttpReponseClientRedirect
     # from django-htmx middleware forces a redirect to the two-factor sign-in page
     if request.htmx:
-        return HttpResponseClientRedirect('two_factor:login', status=403)
+        if request.user.is_authenticated:
+            redirect_url = reverse_lazy("redirect_403")
+            return HttpResponseClientRedirect(redirect_url, status=403)
+        else: 
+            return HttpResponseClientRedirect('two_factor:login', status=403)
     else:
-        return redirect('two_factor:login')
+        if request.user.is_authenticated:
+            return render(
+            request,
+            template_name="epilepsy12/error_pages/rcpch_403.html",
+            context={},
+            status=403,
+        )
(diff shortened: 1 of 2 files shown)


pacharanero:
OK looks great, my only question would be - are there situations in which we DO want a 403 to be expressed and the user sent to the "You need additional privileges" page? For example if they attempt to navigate (by changing the URL in the address bar manually) to a part of the application they are not allowed to see?

The redirect to login screen should only be when the system has auto-logged-out the user.

dc2007git:
> OK looks great, my only question would be - are there situations in which we DO want a 403 to be expressed and the user sent to the "You need additional privileges" page? For example if they attempt to navigate (by changing the URL in the address bar manually) to a part of the application they are not allowed to see?
> 
> The redirect to login screen should only be when the system has auto-logged-out the user.

That makes sense - will take a look and see what I can do.

dc2007git:
@pacharanero this new commit should do exactly what you've suggested

pacharanero (commented):
Added a redirect in `settings.py` so that when clicking on the Log Out button it also takes you back to the login page.

PR #765, opened 2024-01-23
'Other' option in Cause - multiaxial diagnosis

### LINKED ISSUE #723: 'Other' option in Cause - multiaxial diagnosis
Received a phone call and email from a user who has two patients whose causes aren't in the dropdown. 

"1.	Patient has history of neonatal seizures secondary to HIE/Hypoglycaemia. Brian MRI has abnormalities which is consistent with neonatal hypoglycaemia. Now presented with afebrile seizures . Regarding cause there is no other option in structural cause, only all syndromes and genetic cause.
2.	The patient has history of perinatal stroke and abnormal MRI scan. Known hemiplegic CP . Now presented with focal seizures at age 9 years. It is structural cause, but no other option , so could not get 100% data completion in multiaxial diagnosis ( DESCRIBE definition) "

Considering these cases, would it be appropriate for us to add an 'Other' button alongside the dropdown list of causes? 

@nikyraja thoughts?

eatyourpeas:
Fixes #723

### Overview

This adds a work flow largely from the command prompt as a `manage.py` command to update the database with new epilepsy causes. This could be done simply in the django admin, but epilepsy causes we normally pull from our snomed server, so this PR contains a new function(s) to accept a list of snomedct ids, fetch the concepts from our server and store the relevant fields (term and preferredTerm) in the tables.

### Code changes
`migration_0006`: this is where epilepsy causes are seeded. Any new causes are stored here so that any new instances of E12 are run (such as in development) to ensure the causes in live match what is in the new instance

`general_functions/fetch_snomed.py`: 2 functions to make the API call with the SCTID and persist the result. The second accepts a list of sctids and iteratively calls the first. Errors and success are returned as an object and this is used to log to the console.

`management/commands/seed.py`: adds a new mode and parameter. Example: `python manage.py --mode=add_new_epilepsy_causes -sctids 76496008 52767006`
Note accepts a list minimum length of 1. Duplicates are skipped but logged to the console.

I have added a new documentation file to describe the workflow to the docs. I realise now that I forgot to add this to the yml also, so will not show but the text is there.

### Related Issues

#723

### Mentions

### DIFF
--- a/documentation/docs/development/dropdowns.md
+++ b/documentation/docs/development/dropdowns.md
@@ -1,0 +1,26 @@
+---
+title: Dropdowns
+reviewers: Dr Simon Chapman
+---
+
+## Lists
+
+Lists that feed either the toggle buttons or the select dropdowns are either found in the `constants` folder, or are seeded to the database in the `migrations`.
+
+Organisation lists are seeded from `constants` as are the levels of abstraction associated with them (trust, ICB etc.) The Epilepsy causes and comorbidities, by contrast are a SNOMED refset seeded from the RCPCH SNOMED server in migration 0006 and 0009 into EpilepsyCause and ComorbidityList resp. Syndromes and medications are from lists in the `constants` folder.
+
+Since go-live, E12 have wished on the basis of user feedback, wanted to add new items to these lists. The process for adding new items should be:
+
+### EpilepsyCause
+
+1. Epilepsy12 team to supply the SNOMED CT ID (SCTID) of the concept
+2. The development team add the SCTID to the list `extra_concept_ids` in migration 0006 for future seeding from scratch, mostly for development reasons
+3. In the python shell to run the seed function:
+
+```console
+python manage.py --mode=add_new_epilepsy_causes -sctids 764946008 52767006
+```
+
+Note that the function expects a list, even if only one item is supplied.
+
+<!-- There will need to be further documentation added here for new organisations and trust, as well as new comorbidities, and possibly medications and so on. For now, this is the workflow for EpilepsyCauses -->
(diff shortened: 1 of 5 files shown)


eatyourpeas:
Just to demonstrate that a fresh install of E12 adds the additional SNOMED preferredTerms as per #723
To bring staging and live up to date with these the seed function would need to be run in those VPSs.
image

pacharanero:
Do we have a plan as to how we run the seed function in Staging and Live?
I am thinking we possibly need to move seeding out of migrations, since this kind of thing seems to be required fairly frequently.
Each seed function will need to be idempotent (you can run it multiple times safely, and it will not duplicate or over-write changes, only add new entries if they don't already exist)

eatyourpeas:
I have not changed the seed functions with the exception of adding any additional items to the constants and where necessary, making sure those are seeded also on a fresh install. This has meant some additions to migration 006, but this should not affect current state. I have tested it locally.

This PR is for adding extra Epilepsy Causes only and is a `manage.py` command that accepts a list of SNOMED-CT ids. The function looks up the concepts from our SNOMED server and adds them to the EpilepsyCause table. If the item is already there, it will skip it, so in that sense it is idempotent.

The workflow would need to be to run this on the command line in each VPS. For example:
`docker compose exec django python manage.py seed --mode=add_new_epilepsy_causes -sctids 76496008 52767006`
In time I am sure we can find a better way to do this, but we should not be adding epilepsy causes very often, and if so, only in ones and twos.

dc2007git:
This PR also has 34 tests which fail locally, throwing the same ValueError: Missing staticfiles manifest entry for 'hyperscript_0.9.8/hyperscript.min.js' message. Commit which could fix this problem is in #775 , with an update to the hyperscript version

pacharanero (commented):
tested locally including additional command to add SNOMED IDs

PR #770, opened 2024-01-26
Fixes autologout from 30 seconds to 30 minutes

dc2007git:
Converts 30 minutes to seconds (1800) and changes env-template for AUTO_LOGOUT_DELAY_MINUTES accordingly


dc2007git:
@pacharanero looks like Error 137 is getting raised for this check.... We had that before with memory problems. Is this a github/docker compose issue?

anchit-chandran:
# Added to this pull request:

1. Bypass 2fa if `settings.DEBUG is True AND user.is_superuser`
2. Force seeding test to run first (seeds the test_db). for some reason, in the more recent tests, this is not getting captured and being run first which is odd - it's in the top level test directory, so should be the first captured. Hacky workaround was just to run `pytest ...test_meta` first and then run the entire test suite in the workflow file

PR #773, opened 2024-01-27
Remove delete site button

eatyourpeas:
Fixes #772

### Overview

The red delete site button in the registration form is causing confusion and leading to users deleting the site, thinking they are deleting the child. By removing the button for all, this removes this vulnerability.

### Code changes

Remove the delete_site function from `registration_views`, from the urls and from the template.

### Documentation changes (done or required as a result of this PR)

Please describe any changes to documentation here.

### Related Issues

fixes #722

PR #775, opened 2024-01-27
date change event makes ajax call too soon

### LINKED ISSUE #744: Convert logging f-strings to optimised % formatting
Noted by pylinter that using `"%"` formatting is more efficient than using f-strings (related to evaluation of variables delayed with `%` formatting, so especially in the case of costly functions, this is more efficient.

No logging calls use expensive functions so not urgent priority but would be better stylistically.

### LINKED ISSUE #774: date change event makes ajax call too soon
If the date picker is not used, but the date is typed instead, the 1sec time delay is often not enough.

eatyourpeas:
Fixes #774

### Overview

The date field previously had a delay set on the change event, which worked if using the date picker, but not if typing directly into the date field. 

The fix applied here is to add a set button attached to the date field which makes the ajax post request on click

### Code changes

Add a button to the date field partial and move the htmx calls

### Related Issues

~~closes #744~~ closes #774

### Mentions

@pacharanero who first identified the issue

### DIFF
--- a/static/hyperscript_0.9.12/hyperscript.min.js
+++ b/static/hyperscript_0.9.12/hyperscript.min.js
@@ -157,62 +157,62 @@
       }
       while (o < a.length) {
         if (
-          (S() === "-" &&
-            q() === "-" &&
-            (n.isWhitespace(N(2)) || N(2) === "" || N(2) === "-")) ||
-          (S() === "/" &&
-            q() === "/" &&
-            (n.isWhitespace(N(2)) || N(2) === "" || N(2) === "/"))
+          (q() === "-" &&
+            N() === "-" &&
+            (n.isWhitespace(I(2)) || I(2) === "" || I(2) === "-")) ||
+          (q() === "/" &&
+            N() === "/" &&
+            (n.isWhitespace(I(2)) || I(2) === "" || I(2) === "/"))
         ) {
           h();
         } else if (
-          S() === "/" &&
-          q() === "*" &&
-          (n.isWhitespace(N(2)) || N(2) === "" || N(2) === "*")
+          q() === "/" &&
+          N() === "*" &&
+          (n.isWhitespace(I(2)) || I(2) === "" || I(2) === "*")
         ) {
           v();
         } else {
-          if (n.isWhitespace(S())) {
-            r.push(R());
+          if (n.isWhitespace(q())) {
+            r.push(A());
           } else if (
-            !C() &&
-            S() === "." &&
-            (n.isAlpha(q()) || q() === "{" || q() === "-")
+            !R() &&
+            q() === "." &&
+            (n.isAlpha(N()) || N() === "{" || N() === "-")
           ) {
             r.push(d());
-          } else if (!C() && S() === "#" && (n.isAlpha(q()) || q() === "{")) {
+          } else if (!R() && q() === "#" && (n.isAlpha(N()) || N() === "{")) {
             r.push(k());
-          } else if (S() === "[" && q() === "@") {
+          } else if (q() === "[" && N() === "@") {
             r.push(E());
-          } else if (S() === "@") {
+          } else if (q() === "@") {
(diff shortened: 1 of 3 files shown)


eatyourpeas:
I am not sure I understand why tests are failing. They seem to be failing because of previous updates made to the `ETHNICITY` constant which seem to have knock on effects for the tests. I have not touched any of that code here though, and I have taken this issue as branch off what I thought was the latest development.

So to be clear
1. The reasons that tests for this PR are failing are not related to any changes made in the code in this PR
2. Dockerfile has changed at some point the python version to 3.10-slim which is preventing building of dependencies that still use install.py rather than wheels. This is only seen on a fresh install.

pacharanero:
I will revert the changes to the base Python Dockerfile on Monday - I was trying to reduce the size of the Docker images (which are about 2.3Gb each and we have 5 such Python containers running in all environments.). But it has broken dependencies which build from source and so yes it will have to be reverted. We can try other strategies for Docker image size, perhaps we could use a slimmer image for MkDocs, Celery, Celery-beat, and Flower, leaving the heavyweight image for only Django.

pacharanero:
I'm not sure about what is happening to the ETHNICITY constant that is affecting tests, but we have fairly recently made changes there, and also refactored how the ENV vars get supplied into tests, so it could be something going on there.

eatyourpeas:
There are only 2 failing tests:
```
E                   AssertionError: Expected aggregate count for cases_aggregated_by_ethnicity not matching output: aggregate={'ethnicity_display': 'White British', 'ethnicities': 10} should be expected_ethnicity_data={'ethnicity_display': 'White British', 'ethnicities': 11}
E                   assert False
```
Looks to be a mismatch between the number of expected cases seeded in the tests with an ethnicity of White British, compared with the calculated value. 

I have not touched any of that code.

For this PR - I have suggested a new button, here disabled as eligibility criteria for E12 have not been confirmed:
image

Here criteria have been confirmed so the date field is enabled but the button remains disabled as no date has yet been entered:
image

This changes either when the datapicker selects a date, or a date including year have been typed:
image

Once a date has been selected and the set button clicked, the confirmation dialog appears as before with the subsequent workflow which remains unchanged.

dc2007git:
@eatyourpeas the ethnicity changes also seem to be affecting PR #773.

I ran the test suite on this branch locally, and the only changes in my code were to AUTO_LOGOUT_DELAY_MINUTES (setting it to 30 mins instead of 30s) and I had 34 failed permissions_tests - BUT the 2 tests failing on GitHub did pass locally. The errors are all:

ValueError: Missing staticfiles manifest entry for 'hyperscript_0.9.12/hyperscript.min.js'

/usr/local/lib/python3.10/site-packages/django/contrib/staticfiles/storage.py:513: ValueError.

eatyourpeas:
Ah that is because I updated the hyperscript version in the same or maybe its own commit. I refactored the version number in base.html and added the new min.js code to a new folder in `static`. I used hyperscript for the logic disabling the new button in the UI. Hope I have not caused a problem.

dc2007git:
> Ah that is because I updated the hyperscript version in the same or maybe its own commit. I refactored the version number in base.html and added the new min.js code to a new folder in `static`. I used hyperscript for the logic disabling the new button in the UI. Hope I have not caused a problem.

Ah okay, should hopefully not be too difficult to fix. I just ran the test suite again on this PR branch (the last time I did was yesterday) and everything passes locally, so maybe as a matter of priority we merge this PR first in order to resolve the hyperscript versioning issues in other PRs, potentially coming from the rebasing on development that I have performed on these branches. Having said that, I'm not sure why the 34 tests failed the first time I ran them yesterday on this branch, and then passed today, when the hyperscript update commit was present in this branch the whole time. What do you think @pacharanero @eatyourpeas

pacharanero (commented):
tested locally, checked functionality, tests passing

PR #776, opened 2024-01-28
case pagination not working if filter applied

### LINKED ISSUE #766: case pagination not working if filter applied
An observation made by @AmaniKrayemRCPCH 
On typing in the search box in the case list, the search filter is lost/removed on clicking the previous and next pagination buttons.

eatyourpeas:
also adds pagination counter
Fixes #766

### Overview

A bug noticed by E12 that on filtering cases using the search box, the filter term is lost on clicking on the next and previous page buttons.

The reason for this is that the filter term was not added to the URL and therefore the filtering was lost on table refresh.

### Code changes

The filter term passed to `case_list` (via `cases`) on keyup from the search box is passed back to that template in the context. This allows the template to add it to the URL on clicking the next and previous page buttons.

For clarity, an extra label has been added to the footer of the cases table to notify the user how many pages are viewable, and the number of the current page.

No implications for documentation

### Related Issues

Fixes #766

### DIFF
--- a/epilepsy12/templatetags/epilepsy12_template_tags.py
+++ b/epilepsy12/templatetags/epilepsy12_template_tags.py
@@ -479,13 +479,24 @@ def no_eligible_cases(aggregation_model, kpi_name: str):
 # the list of which can be found in 'capitalised_words' below
 @register.filter
 def capitalise_org_names(organisation_name):
-    capitalised_words = ['Ii', 'Rbh', 'Nhs', 'Cdc', '(Hq)', '(Epma)', '(Epact)', 'Gstt', 'Qmc', 'Ctr']
+    capitalised_words = [
+        "Ii",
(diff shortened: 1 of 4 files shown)


eatyourpeas:
Adds a pagination label to the footer of the table as well as fixes the bug.
image

dc2007git (commented):
Happy to merge this once Marcus has OK'ed it. Some nice functional improvements, so thanks @eatyourpeas!

Am going to just restyle the page count so it lies below the Next/Previous buttons too

Edit 15:51 - tests have just finished locally running and 34 fail, relating to ValueError: Missing staticfiles manifest entry for 'hyperscript_0.9.8/hyperscript.min.js'

@eatyourpeas it looks like there are changes to hyperscript in #775 . Is this something related to that?

PR #815, opened 2024-02-29
Eatyourpeas/issue813

### LINKED ISSUE #813: User with missing registration
A Lead user contacted me saying she's trying to delete a user from the platform but can't.

I tried to do it myself and, while i could click the delete button and the pop-up would appear, i couldn't actually delete them. They would remain on the list. 

I then went into the admin panel to delete them, while the user was within the list of Epilepsy12 users, there was no registration for this user. 

Instead, when I try to delete the user I'm met with this message, asking me to delete a different person's registration:

Image

I can't delete this user as they have no registration and it's asking me to delete someone else's registration. For the time being, I've removed their permissions.

eatyourpeas:
### Overview

Bug fix
E12 users identified that deleting a user who had previously scored a child threw a referential integrity error, as the registration for that child contains the foreign key of the user in the `updated_by` field.
Users can therefore only be deleted if they have not yet created any registrations, clearly a problem.
This has been fixed by changing the language around deletion to deactivation, and setting the `is_active` flag in the user model to False.
This shows in the user table as red cross.
This PR also contains a fix in the template for the popup as this was not rendering due to misnaming error of the element id.

### Code changes

changes to 3 files:
`user_management_views`: 2 functions affected, one a callback from the delete button click in the form template (`admin_create_user`, the other an htmx callback from the delete button in the `epilepsy12_user_table`. Both changes set the is_active flag to True and remove the delete.
Refactoring in the templates changes the wording around deletion to deactivation and fixes the popup error.

No documentation changes made.

### Related Issues

closes #813

### DIFF
--- a/epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py
+++ b/epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py
@@ -104,7 +104,9 @@ from epilepsy12.models import (
     Medicine,
     ComorbidityList,
 )
-from epilepsy12.tests.view_tests.permissions_tests.perm_tests_utils import twofactor_signin
+from epilepsy12.tests.view_tests.permissions_tests.perm_tests_utils import (
+    twofactor_signin,
+)
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


mbarton:
Tested this on the branch:

- Create a superuser
- Use the superuser to create a normal user
- Log in as the normal user
- Create a child and add an episode
  - It might have been enough just to create a child, but the updated_by on the child didn't show any value in the admin

Before the change the user remained in the list, after the change they show as deactivated

epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py
```diff
@@ -172,35 +174,6 @@ def test_user_delete_success(
             response.status_code == HTTPStatus.OK
         ), f"{test_user.first_name} (from {test_user.organisation_employer}) requested `{url}` for User from {TEST_USER_ORGANISATION}. Has groups: {test_user.groups.all()}. Expected 200 response status code, received {response.status_code}"
 
-        # Additional test for deleting users outside of own Trust
```
mbarton, line 175: At first glance this test would still be useful even after this PR?

When I put it back I get an integrity error:

```
FAILED epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py::test_user_delete_success - django.db.utils.IntegrityError: duplicate key value violates unique constraint "epilepsy12_epilepsy12user_email_key"
```

Do we know why this fails? I am able to deactivate a user outside of my own organisation in my local instance, although that may be because I'm a super user

epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py
mbarton, line 1: I guess these specs should say deactivated now? and remove the one for the deleted test (which I think is line 12?)

eatyourpeas: Yes good point. I will refactor the advice

epilepsy12/tests/view_tests/permissions_tests/test_permissions_delete.py
eatyourpeas: It fails because after deleting the user, it is reseeding the database with the same user for the next test. Of course in this instance the user has not been deleted, merely deactivated so it throws a database error when trying to create again. I removed the test because I was not clear that it added much, and actually the tests are testing for a 200 or 403 on accessing the function, not actually testing if the user has been removed.

The reason you can edit/update/delete any user is because you have superuser access as you say. If you were to create a lead clinician account and login with that you should be constrained to your own organisation view, and if you try and login and then do a post request through postman (like how many neurologists will do that? but anyway) you should find it throw a 403.

Possibly to close this issue then I might create a test that checks the model rather than the response?

eatyourpeas: [cf655a2](https://github.com/rcpch/rcpch-audit-engine/pull/815/commits/cf655a2e577d3617b5b8622e37f8ca2243c813b1) does this - now deactivation of a user is tested in 2 ways: first that the correct access to the function is given to the correct users, judged by whether a 200 or a 403 is returned; second that any deactivated user is reflected as such in the database.

@mbarton I would be grateful if you are happy with this and the solution, perhaps then merge the request? Thanks again.

mbarton: Sounds good, thanks for the detailed explanation and fix! Great to have this discussion in the PR in case any future devs are wondering :)

PR #821, opened 2024-03-05
Speed up seeding data for local development

mbarton:
4 times faster seeding data by removing API lookups performed per case or registration.

I also added a script to connect to the local db (useful whilst testing this PR).

I also added Kings to the list of seed organisations as that's the one you're sent to when you first log in as a superuser whilst developing. I got confused where my data was until I realised I needed to switch org!

**Before**

```
 Seeding 1000 fictional cases... 
Creating 142 Cases in ADDENBROOKE'S HOSPITAL
Creating 142 Cases in AIREDALE GENERAL HOSPITAL
Creating 142 Cases in ALDER HEY CHILDREN'S HOSPITAL
Creating 142 Cases in BRONGLAIS GENERAL HOSPITAL
Creating 142 Cases in CHELSEA & WESTMINSTER HOSPITAL
Creating 142 Cases in CHEPSTOW COMMUNITY HOSPITAL
Creating 142 Cases in YSBYTY YSTRAD FAWR
docker compose exec django python manage.py seed --mode=seed_registrations --cohort 6 $*
register cases in audit and complete all fields with random answers...
 Registering fictional cases in Epilepsy12... 
 Completing all the Epilepsy12 fields for the fictional cases... 


real    12m21.293s
user    0m0.189s
sys     0m0.064s
```

**After**

```
 Seeding 1000 fictional cases... 
Creating 125 Cases in ADDENBROOKE'S HOSPITAL
Creating 125 Cases in AIREDALE GENERAL HOSPITAL
Creating 125 Cases in ALDER HEY CHILDREN'S HOSPITAL
Creating 125 Cases in BRONGLAIS GENERAL HOSPITAL
Creating 125 Cases in CHELSEA & WESTMINSTER HOSPITAL
Creating 125 Cases in CHEPSTOW COMMUNITY HOSPITAL
Creating 125 Cases in KING'S COLLEGE HOSPITAL (DENMARK HILL)
Creating 125 Cases in YSBYTY YSTRAD FAWR
docker compose exec django python manage.py seed --mode=seed_registrations --cohort 6 $*
register cases in audit and complete all fields with random answers...
 Registering fictional cases in Epilepsy12... 
 Completing all the Epilepsy12 fields for the fictional cases... 


real    3m10.061s
user    0m0.163s
sys     0m0.041s
```

### DIFF
--- a/epilepsy12/models_folder/case.py
+++ b/epilepsy12/models_folder/case.py
@@ -143,7 +143,7 @@ class Case(TimeStampAbstractBaseClass, UserStampAbstractBaseClass, HelpTextMixin
     def save(self, *args, **kwargs) -> None:
         # calculate the index of multiple deprivation quintile if the postcode is present
         # Skips the calculation if the postcode is on the 'unknown' list
-        if self.postcode:
+        if self.postcode and not self.index_of_multiple_deprivation_quintile:
             if str(self.postcode).replace(" ", "") not in UNKNOWN_POSTCODES_NO_SPACES:
                 try:
                     self.index_of_multiple_deprivation_quintile = imd_for_postcode(
--- a/epilepsy12/management/commands/create_e12_records.py
+++ b/epilepsy12/management/commands/create_e12_records.py
@@ -1,5 +1,5 @@
 # python dependencies
-from random import randint, getrandbits, choice
+from random import randint, getrandbits, choice, choices
 from dateutil.relativedelta import relativedelta
 from datetime import date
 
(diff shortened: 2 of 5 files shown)


epilepsy12/models_folder/case.py
```diff
@@ -143,7 +143,7 @@ def age(self, today_date=date.today()):
     def save(self, *args, **kwargs) -> None:
         # calculate the index of multiple deprivation quintile if the postcode is present
         # Skips the calculation if the postcode is on the 'unknown' list
-        if self.postcode:
+        if self.postcode and not self.index_of_multiple_deprivation_quintile:
```
mbarton, line 146: This is a behaviour change: before we would look up this data on every save. But I don't think anything relied on that @eatyourpeas?

eatyourpeas: This looks to be something I must have missed. Thanks for identifying. Yes, there is no need to make an API call if the IMD quintile has already been calculated and saved in the record.

Having said that, I have found an edge case where a child's record can be created with a postcode but then edited to be postcode unknown but the imd field does not get set to None. This feels to me a bug. Also, I notice as minor UI thing that if you click the 'no postcode' button, the three no postcode options are set to none (appropriately), but the save button is not disabled, and really should be until a selection has been made. We could fix as part of this PR, or create a new issue as it is really, but close this. I am happy with whichever approach you feel works best.

eatyourpeas:
I hope you don't mind the cheeky additional commit. I nearly had a heart attack when I saw having seeded my local instance with 10000 kids that the total registrations in the case table was correct (10,000) but the summary page showed considerably lower totals:
image
I thought OMG there is a mistake in the aggregations...
And then when I dug deeper I saw that the seed function was seeding across cohort 6 and 7, and this was because the upper limit of the random date was set to today, not the end date of the cohort. So actually while it was registering some kids in cohort 6, it was about 20% of the time seeding kids in the current cohort, 7.
So I fixed this and now the summary page reflects what it is expected.

mbarton:
Ah great - I had also wondered why the numbers were down.

For the postcode issues let me look today - if they were caused by the changes in this PR I’ll fix them here otherwise open a new one

mbarton:
I think seeding across multiple cohorts will be useful but opt in - can be another PR

epilepsy12/models_folder/case.py
mbarton: Cool let's fix them in separate PRs. Even though we're not doing continuous deployment on this project at the moment I think in general raising small focused PRs will be easier to manage both now and in the future.

I've raised #824 and #825 to track those.

@eatyourpeas just checking you're happy for me to merge this PR? :)

eatyourpeas:
I agree seeding across multiple cohorts a good idea. I would like as part of it to add a select in the UI to allow the user to toggle between cohorts initially as this would not be a big feature but would probably be big impact in December once the next cohort starts recruiting, and in future possible produce graphs of the individual measures from 2 or more cohorts side by side to allow teams to show the impact of any improvements they might have put in place.

Issue #822, opened 2024-03-05
Users are logged out on every restart


mbarton:
This is annoying during development and could cause users to lose progress in the future if we move to continuous deployment (although the app is designed to save progress frequently so in practice it's probably only an issue for developers)

mbarton:
Fixed in https://github.com/rcpch/rcpch-audit-engine/pull/829

PR #828, opened 2024-03-07
Remove Celery, Redis and Flower

mbarton:
We have benchmarked our aggregation queries against 10,000 fake cases. Calculating them at all levels only takes 1 second.

As such we've decided to remove the current async task infrastructure. It's acceptable to wait 1 second to refresh the KPIs in the UI. It also allows us to simplify the dev ops setup in the future, potentially moving back to Azure app service as docker compose will no longer be required.

We will need asynchronous tasks or scheduling in the future and we'll review our tech choices again then based on priorities and needs.

### DIFF
--- a/envs/env-template
+++ b/envs/env-template
@@ -27,7 +27,7 @@ DJANGO_ALLOWED_HOSTS="${SITE_DOMAIN}"
 DJANGO_CSRF_TRUSTED_ORIGINS="https://${SITE_DOMAIN}"
 DJANGO_STARTUP_COMMAND="python manage.py runserver 0.0.0.0:8000" # for local development with auto-reload
 # DJANGO_STARTUP_COMMAND="gunicorn --bind=0.0.0.0:8000 --timeout 600 rcpch-audit-engine.wsgi" # for live deployment
-E12_SECRET_KEY= # secret key for Django
+DJANGO_SECRET_KEY= # secret key for Django
 SITE_CONTACT_EMAIL="sitecontactemail@example.com" # email address for site contact
 
 # DJANGO LOGGING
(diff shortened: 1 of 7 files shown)


epilepsy12/general_functions/email.py
```diff
@@ -0,0 +1,18 @@
...
+from django.utils.html import strip_tags
+from django.core.mail import send_mail
+from django.conf import settings
+
+
+def send_email_to_recipients(recipients: list, subject: str, message: str):
+    """
+    Sends emails
+    """
+    send_mail(
```
mbarton: After discussing during in person code review we decided to remove the try/catch that was in the old version of this function as it would mask issues if they were to occur in production and we are confident that sensitive error data won't be revealed to users in production

pacharanero:
when this PR is merged we can:

* remove unused `CELERY*` and `FLOWER*` env vars from GitHub's `ENVIRONMENT` secret (requires an authenticated, permissioned gh CLI tool already set up so I'm happy to do this)

* remind all devs to remove unused env vars from their **local** .env file(s) cc @dc2007git @anchit-chandran @eatyourpeas @mbarton 

* ~~separately, need to rename E12_SECRET_KEY to DJANGO_SECRET_KEY and then you won't be logged out every time (see #822)~~ this is now done, added to this PR

pacharanero (approved):

eatyourpeas (commented):
So this is all looking good but I am running into some small issues before I close this. The containers build nicely, the database seeds and the aggregations run quickly as above. The tests equally are all passing.

The other async function was emailing and account creation seems to send the email to the console in dev correctly. But when trying to logout and log in to the new user account I get a 405 error and cannot log out.

image

To add though, the transfer lead site emails are working as expected.

**EDIT**
So actually this is a red herring and reinforces once more the importance of freezing the versions of our dependencies. Django 4.1 has deprecated log out on GET request as a security feature - see [release notes](https://docs.djangoproject.com/en/5.0/releases/4.1/#log-out-via-get). I have updated the nav bar to accept a POST request instead and this is now working for me now.

mbarton:
@eatyourpeas no worries, I agree about freezing our dependencies. I'll raise a PR for that and include changing the logout

Issue #791, opened 2024-02-11
KPI Data Export Specification


pacharanero:
[DataExportTemplate.xlsx](https://github.com/rcpch/rcpch-audit-engine/files/14232036/DataExportTemplate.xlsx)

The attached XLSX document is a template for how the E12 team would like to receive the exported KPI data. At present this is the most pressing priority in terms of data export.

@AmaniKrayemRCPCH @sairapons @nikyraja Is this format your _preferred_ format or just the format you have been used to because of Netsolving? I ask because this is an opportunity to improve the export format to make any onward processing steps easier. We will be using code to collect the KPI data and process it into a file you can download, so we have options for how that file will look.

Some questions to get us started:

* There are 8 Sheets within the XLSX file, is it best if these are all in one XSLX at the point of export, or would eight separate, semantically-named CSV files be better? Or is this export file already _exactly_ what you want to be able to download?

* In terms of implementation, I envision a page, accessible only to RCPCH Audit Team members, with a button called 'KPI Download'. When this button is pressed, the code will query the latest pre-saved version of the KPI calculation numbers, format it as a file, and let you download it. There should be some explanatory text on that page as well. There should be a 'KPI download' section in the Documentation. Does this sound right to your team?

I think in order to get this nailed we should do some preliminary coding work to get a 'rough' export button and file working, and then do some 1:1 work with an analyst and a programmer to get the details nailed exactly, for example ordering of rows, ordering of columns, Sheet naming and ordering etc.

AmaniKrayemRCPCH:
[DataExportTemplate.xlsx](https://github.com/rcpch/rcpch-audit-engine/files/14239851/DataExportTemplate.xlsx)

@sairapons has provided an updated version of the template. 

This is the preferred format for the data, as it is the format required for updating the Power BI dashboard. It's especially important to keep the column titles the same. 

I believe the 8 sheet .xlsx file is the preferred format, and the 'KPI download' button sounds good!

pacharanero:
@dc2007git @pacharanero this is ready for working on next week.

pacharanero:
* [x] button to take user to KPI download page
* [x] KPI download page
* [ ] Should probably say when KPIs were last calculated (ie how old the data is) - (Could offer user a 'recompute KPIs now' button???)
* [x] Download KPIs button
* [x] KPI data model to CSV logic/code (using Pandas?)

[...]

PR #831, opened 2024-03-07
filter_completed_cases_not_all_cases

### LINKED ISSUE #830: Aggregations run on all cases, both incomplete and complete
The filter query prior to aggregating all KPIs in pulling out all cases registered in a given cohort, *including* those who are not fully scored. 
What is happening
100 cases registered in cohort 6 in organisation X, 95 have been completed, 5 have some unscored measures and are incomplete
Organisation X summary table should show total 95 cases but actually shows 100 cases

eatyourpeas:
### Overview

The filter query prior to running aggregation maths is selecting all cases registered in a given cohort at a given level of abstraction, and should be selecting only those in a given cohort at a given level of abstraction who have completed all the measures.

This is a small fix which corrects this.

### Code changes

A single change to the select query in the `aggregate_by` file, adding the additional terms to the query.
This does not affect the tests, but we should probably include tests at a later date test this.

No docs to update

### Related Issues

closes #830 has implications for #791

### Mentions
@dc2007git

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -23,6 +23,8 @@ logger = logging.getLogger(__name__)
 """
 Reporting
 """
+
+
 def cases_aggregated_by_sex(selected_organisation):
     # aggregate queries on trust level cases
 
@@ -506,9 +508,16 @@ def update_all_kpi_agg_models(
     Case = apps.get_model("epilepsy12", "Case")
 
     all_cases = Case.objects.filter(
-        site__site_is_actively_involved_in_epilepsy_care=True,
-        site__site_is_primary_centre_of_epilepsy_care=True,
-        registration__cohort=cohort,
+        Q(site__site_is_actively_involved_in_epilepsy_care=True)
+        & Q(site__site_is_primary_centre_of_epilepsy_care=True)
+        & Q(registration__cohort=cohort)
+        & Q(registration__audit_progress__registration_complete=True)
+        & Q(registration__audit_progress__first_paediatric_assessment_complete=True)
+        & Q(registration__audit_progress__assessment_complete=True)
+        & Q(registration__audit_progress__epilepsy_context_complete=True)
+        & Q(registration__audit_progress__multiaxial_diagnosis_complete=True)
+        & Q(registration__audit_progress__investigations_complete=True)
+        & Q(registration__audit_progress__management_complete=True)
     )
 
     abstraction_levels = EnumAbstractionLevel if abstractions == "all" else abstractions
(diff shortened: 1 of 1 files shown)


dc2007git (commented):
Looks good Simon. So the problem is that cases are appearing as completed in the dashboard even if they're only partially completed? 

As I understand it, KPIs are calculated even on those who have not fully completed their audit - we use their partially completed data in the KPI calculation. Would filtering the cases with the & operator not remove data from those cases which are partially completed? It's obviously what we want for the dashboard but just making sure for the KPI aggregation functionality as well

eatyourpeas:
Thank you Danny. In that case I have possibly misunderstand a basic step in the calculation. For sure @nikyraja was clear from the start that E12 team wanted to use all the data for their calculations, and I think the implication was that it did not matter if cases had not been completed, but I do not remember that being explicitly said. Nikky could you please confirm that you want all cases in the calculations, even if they have not completed filling the audit?

If that is the case, then I will close this PR and we need take no further action. Just to be aware though, there are 2 sets of aggregations: those that power the tables in the dashboard, and those that power the pie charts. The pie chart select queries **do** filter out incomplete cases, so possibly we should raise a separate issue if Nikky confirms above to ensure we are consistent and remove this higher standard in that query.

nikyraja:
Thanks @dc2007git and @eatyourpeas.

We've reflected on this and have reached the decision that it is okay to include CYP with partially completed records in the KPI calculations. Ie, a record doesn't have to be complete for it to be included in the trust, ICB, region etc aggregates in the dashboard.

**However, it is important that the aggregated KPI calculations in the dashboard only include CYP who have completed a full year of care, ie. one year has passed from their first paediatric assessment date.**   Apologies that I had not explicitly said this before.

Our annual analyses only include those CYP who have completed a full year (the deadline makes this true for everyone in the cohort) so it is important that the dashboard follows the same methodology. Otherwise, we will have instances where CYP are included in the live dashboards and the results are therefore different to our published outputs. This would cause confusion and effect the confidence users have in our data. 

The KPI info in the individual patient records (image attached) do not need this restriction to 1 year though. These can indicate for example that the KPI has not been met yet, and this gives clinicians time to rectify this before the year is completed, and promotes prospective data entry.
image

We can discuss this on Friday perhaps?

eatyourpeas:
Thanks @nikyraja that is really helpful
The KPI aggregations accept a cohort as a parameter, so I am assuming that all those in cohort 6 have already completed their first year of care.

Those that appear in the dashboard likewise only reflect those that are in cohort 6, whether they have completed the audit measures or not. An implementation in the future that is in the issues is to add a dropdown that allows users to select which cohort to filter the results to. Your KPI CSV downloads which have been built by @dc2007git should only include children in cohort 6.

nikyraja:
Cohort 6 includes patients who had a first assessment between 1 December 2022 and 30 November 2023. Patients will then complete their first year of care between 1 December 2023 to 30 November 2024. 

After the 30/11/24, all patients in cohort 6 will have completed the first year of care, however at this point only about half would have. Does that make sense?

Is there a way to build in this filter into the dashboard calculations and therefore the KPI cvs download? @dc2007git @eatyourpeas

eatyourpeas:
Oh I see.
Yes, ok. Let me look at that.

eatyourpeas:
I think the conversation has moved on with this issue from what was originally proposed in the PR.
I will therefore close this PR without merging anything and open a new issue with your comment above Niky.

PR #835, opened 2024-03-08
Removes broken has_group filter

### LINKED ISSUE #807: request.user|has_group filter is broken
I've spotted a few instances throughout the code of request.user|has_group being used as a filter to conditionally render some elements depending on the user's group. 

It seems as though there is a bug with this filter however, as in all places where it is called with the group set to 'epilepsy12_audit_team_full_access', it fails to show its contents to the screen.

The fix to this is simple - we use `request.user.is_<role>` as is used (and works) elsewhere in the codebase.

dc2007git:
### Overview

In order to conditionally render specific areas of html, the has_group filter was designed. However, it does not work and we prefer to use granular permissions by checking the user's individual permissions for each use case instead.

@mbarton there is another mention of the has_group filter in the code, but as part of the KPI export function branch last week I removed it and replaced it with a granular permission - so it is still here in this branch. What is best practice for this - should I remove and replace it in this branch also or just wait until the KPI export branch is merged?

### Code changes

Remove has_group filter. There are so far still 2 instances in the code but are there for the reason described above. 

### Related Issues

Closes #807


mbarton (commented):
Thanks @dc2007git! I would say get the KPI work done first and then update this PR and merge separately - as this functionality is broken but not presenting any issues atm

dc2007git:
> Thanks @dc2007git! I would say get the KPI work done first and then update this PR and merge separately - as this functionality is broken but not presenting any issues atm

Yeah tests are failing on this branch! KPI work is pretty much done, showing the E12 team today then hopefully we can merge.

eatyourpeas:
The tests broke because you removed the template tag but did not remove from the templates. Like you said, there are 2 instances. I have refactored the first to use a permission, and removed the second as that was the 'Frida button'. I think that needs deprecating - it was a button to allow the team to download the tables as csv files, but given we are implementing a new solution I am deprecating it here.

PR #847, opened 2024-03-15
Sets autoescape to on for password reset page

### LINKED ISSUE #767: Password reset in local development not working
Upon entering my email address in the password reset button in local development (even for the first time), I get raised the following error:
image

This is not the case in e12.rcpch.ac.uk - it is still fully functional there. However, it does happen even when running the live branch locally.

Can anyone replicate this problem while on the HEAD of development? @pacharanero @eatyourpeas @anchit-chandran

dc2007git:
### Overview

Provides a fix to the password reset field being broken in local development.

This change fixes the breaking, but also provides an extra layer of security against XSS attacks by autoescaping any html characters.

### Code changes

Changes password_reset_email.html, sets autoescape to on.

### Related Issues

Closes #767 

### Mentions

@mbarton

### DIFF
--- a/templates/registration/password_reset_email.html
+++ b/templates/registration/password_reset_email.html
@@ -1,6 +1,6 @@
 {% load epilepsy12_template_tags %}
 {% load static %}
-{% autoescape off %}
+{% autoescape on %}
 <div class="indent" style="display: relative; width: 80%; margin: auto;">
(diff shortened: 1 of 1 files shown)


mbarton:
Sorry for the delay getting to this one - I just need to run it locally and check what Simon mentioned about the email formatting

PR #857, opened 2024-03-29
Eatyourpeas/issue845

eatyourpeas:
### Overview

Adds epilepsy cause 1196837008 to migration_006
Updates docs


eatyourpeas:
Actually development and staging have slightly diverged so this change is actually in staging but not development.
I am inclined to delete this PR and branch therefore.

Issue #915, opened 2024-04-26
Reorder columns in KPI report


mbarton:
Spun out from https://github.com/rcpch/rcpch-audit-engine/issues/895#issuecomment-2075179987

    in HBT_level = HBT, Measure, Percentage, Numerator, Denominator
    in ICB_level = ICB, Measure, Percentage, Numerator, Denominator
    in NHSregion_level = NHSregion, Measure, Percentage, Numerator, Denominator
    In Network_level = Network, Measure, Percentage, Numerator, Denominator
    in National_level = uk, Measure, Percentage, Numerator, Denominator

Issue #895, opened 2024-04-23
KPI data download for April


AmaniKrayemRCPCH:
We're hoping to update our Power BI dashboard on Wednesday 1st May. 
Can we be sent the KPI data download for May, as outlined in https://github.com/rcpch/rcpch-audit-engine/issues/791

Trust mergers don't need to be actioned by then, as we can action these before publication.

mbarton:
I'll keep a list of issues that need to be fixed here as I or others raise them:

**Blockers**

- https://github.com/rcpch/rcpch-audit-engine/issues/899
- https://github.com/rcpch/rcpch-audit-engine/issues/900
- https://github.com/rcpch/rcpch-audit-engine/issues/901
- https://github.com/rcpch/rcpch-audit-engine/issues/910
- https://github.com/rcpch/rcpch-audit-engine/issues/907
- https://github.com/rcpch/rcpch-audit-engine/issues/912
- https://github.com/rcpch/rcpch-audit-engine/issues/924
- https://github.com/rcpch/rcpch-audit-engine/issues/931
- https://github.com/rcpch/rcpch-audit-engine/issues/934

**Nice to have**

- https://github.com/rcpch/rcpch-audit-engine/issues/913
- https://github.com/rcpch/rcpch-audit-engine/issues/914
- https://github.com/rcpch/rcpch-audit-engine/issues/915

AmaniKrayemRCPCH:
The below issue may affect this

*  #901

mbarton:
Thanks @AmaniKrayemRCPCH! @eatyourpeas and I will review that issue asap

mbarton:
@AmaniKrayemRCPCH the download KPI button is now live in staging. It's a little confusingly placed on the organisation page even though it gives you data across all orgs but is only visible to RCPCH admins:

<img width="505" alt="Capture" src="https://github.com/rcpch/rcpch-audit-engine/assets/395805/a813d7eb-e25b-4679-b30e-b370e35185e8">

Could you try and it let me know if there's anything obviously missing. We'll investigate #901 before deploying this to production and of course can do some more rounds if the data isn't correct there, given staging has very limited data.

AmaniKrayemRCPCH:
Thanks @mbarton! 

I've put some comments below in order of priority. I believe we can correct all of these ourselves after downloading, but it would be helpful if these were resolved (particularly the first one).

[...]

PR #920, opened 2024-04-27
Derive list of trusts for KPI report from organisations

### LINKED ISSUE #912: Remove irrelevant trusts from KPI report
Spin off from https://github.com/rcpch/rcpch-audit-engine/issues/895#issuecomment-2075179987

> There are some extra Trusts in there that aren't part of the audit e.g. South East Coast Ambulance Service NHS Foundation Trust. There's also a number of closed Trusts. This was raised in https://github.com/rcpch/rcpch-audit-engine/issues/791. I believe Danny said in a meeting that this was resolved. If it helps, it seems that none of the extra trusts have an associated ICB.

@AmaniKrayemRCPCH has sent me the list of trusts participating in the audit so we can cross reference, although that does not include some mergers that are [EDIT] already actioned in the platform.

I think the easiest way to proceed for now is to compare that list with ours and remove any extra. The removal is a three step process, remove from the seeding code and then remove manually from the Django admin in staging then live. We must double check no data has been entered under those trusts, even though the Django relation management should stop us deleting them if that's the case.

We need a longer term solution for actioning trust mergers but this issue is simply scoped to what we need to do to get the KPI report out of the door.

mbarton:
This automatically excludes trusts that are not relevant to the audit.

Fixes #912

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -907,7 +907,7 @@ def create_KPI_aggregation_dataframe(
         )
 
     if model_aggregation_2:
-        if constants_list2 == TRUSTS:
+        if constants_list2:
             for i, body in enumerate(constants_list2):
                 key = body["ods_code"]
                 uid = i + 1
@@ -934,7 +934,7 @@ def create_KPI_aggregation_dataframe(
         title = "Network"
 
     # Group KPIs by Trust, and add to dataframe - ie collect all KPIs for a specific trust, then add to dataframe
-    if constants_list2 == TRUSTS:
+    if constants_list2:
         for key in objects:
             object = objects[key]
             for index, kpi in enumerate(measures):
@@ -1015,7 +1015,7 @@ def create_reference_dataframe(trusts, health_boards, networks, icbs):
     # For the English Trusts - create row with HBT, HBT_name, Network, Country, UK, NHSregion and ICB as the column names
     for trust in trusts:
         ods_code = trust["ods_code"]
-        trust_name = trust["trust_name"]
+        trust_name = trust["name"]
         network_name = ""
         region = ""
         icb = ""
--- a/s/django-shell
+++ b/s/django-shell
@@ -1,0 +1,3 @@
+#!/bin/bash -e
+
+docker compose run -it django python manage.py shell
--- a/epilepsy12/kpi.py
+++ b/epilepsy12/kpi.py
@@ -6,8 +6,6 @@ import pandas as pd
 
 # E12 Imports
 from epilepsy12.constants import (
-    EnumAbstractionLevel,
-    TRUSTS,
     LOCAL_HEALTH_BOARDS,
     INTEGRATED_CARE_BOARDS,
     NHS_ENGLAND_REGIONS,
@@ -20,6 +18,7 @@ from epilepsy12.common_view_functions.aggregate_by import (
     create_reference_dataframe,
     create_kpi_report_row,
 )
+from epilepsy12.models import (Organisation, Trust)
 
 
 def download_kpi_summary_as_csv(cohort):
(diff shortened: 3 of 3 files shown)


epilepsy12/common_view_functions/aggregate_by.py
```diff
@@ -907,7 +907,7 @@ def create_KPI_aggregation_dataframe(
         )
 
     if model_aggregation_2:
-        if constants_list2 == TRUSTS:
```
mbarton, line 910: This check was redundant - we only accessed `constants_list2` if this was true.

I think it would be clearer to remove the checks against `constants_list1` too but that's for a future PR

s/django-shell
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
mbarton, line 1: I found this useful so included it in the PR

eatyourpeas (commented):
This all looks good - would you like me to merge or just review? 

Can I just check a few items?

In the spreadsheet generated at country level I am seeing:
1. Paediatrician with expertise
1. Paediatrician with expertise
2. Epilepsy specialist nurse
2. Epilepsy specialist nurse
3a. Tertiary involvement
3a. Tertiary involvement
3b. Epilepsy surgery referral
3b. Epilepsy surgery referral
4. ECG
4. ECG
5. MRI
5. MRI
6. Assessment of mental health issues
6. Assessment of mental health issues
7. Mental health support
7. Mental health support
8. Sodium valproate
8. Sodium valproate
9a. Comprehensive care planning agreement
9a. Comprehensive care planning agreement
9b. Comprehensive care planning content
9b. Comprehensive care planning content
10. School Individual Health Care Plan
10. School Individual Health Care Plan

That is, 2 rows for each measure - is that one set of measures for each country (England and Wales)? How do we know which country is which?

At trust level, the ODS code is present, but no name column? Is that correct? Makes it a little harder to review the PR since this is to check the TRUST list is filtering the correct trusts. I am not sure if it is right - at ~1800 rows, with 12 scoreable measures = 150 trusts/local health boards. We have 242 trusts in our list and 7 local health boards in Wales, so more like 250,  but of course many of these are not used in E12. I am not sure how many trusts finally E12 are asking us to use


This is true for epilepsy network also - only the code, not the name.

At all other levels, the results are ordered by measure, not ICB, is that right? I can see arguments for doing it both ways?

I think probably this is about useability of the data, rather than any intrinsic problem with the PR itself (which is readable and elegant).

eatyourpeas (approved):
Sorry, have put my comments in the PR discussion, not here. Happy to approve.

mbarton:
Thanks for the comments - I think it's important to spot check this data as much as we can.

> In the spreadsheet generated at country level I am seeing ... 2 rows for each measure - is that one set of measures for each country (England and Wales)? How do we know which country is which?

Yes that looks like a mistake. I've raised an issue for it: https://github.com/rcpch/rcpch-audit-engine/issues/924

> At trust level, the ODS code is present, but no name column? Is that correct?
> ...
> I am not sure how many trusts finally E12 are asking us to use

I'm not sure about which columns are required. The template spreadsheet in #791 has `sitename`, `network_name`, `network_acr`, `country`, `nhsregion` and `icb`, none of which we have in our current export. But Amani [commented here](https://github.com/rcpch/rcpch-audit-engine/issues/895#issuecomment-2075179987) to re-arrange the columns only including `HBT`, `Measure`, `Percentage`, `Numerator`, `Denominator` as we do currently. I think we will need to confirm again with the team.

Either way I have spot checked the ODS codes in the report against the list of relevant trusts they provided and I don't think we are missing any or have any additional ones. There are some codes in their list that our not in the report but those are the mergers we have already actioned on the platform.

> At all other levels, the results are ordered by measure, not ICB, is that right? I can see arguments for doing it both ways?

Yes there's https://github.com/rcpch/rcpch-audit-engine/issues/915 to re-order the columns

PR #925, opened 2024-04-29
Adds Country column to Country - Sheet 2

### LINKED ISSUE #924: Missing country column in KPI export

dc2007git:
### Overview

Adds the column 'Country' to sheet number 2 - 'Country'

### Code changes

Adds a line to the create_kpi_report_row function to render either 'England' or 'Wales' as the country and add it to the ret dictionary object

N/A

### Related Issues

Closes #924

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -1,6 +1,7 @@
 from typing import Literal, Union
 from datetime import date
 import logging
+from collections import OrderedDict
 
 # Django imports
 from django.apps import apps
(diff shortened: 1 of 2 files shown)


mbarton:
This works but I'd like to avoid putting sheet specific modifications low down in the call stack where we can. It makes it harder to understand what the low level function is doing without holding all the calling code in your head at the same time.

Could we add the extra column in `kpi.py` only for sheet 1? eg

[kpi.py:94](https://github.com/rcpch/rcpch-audit-engine/blob/c7bcb6956c523817c59f6e4ba3025e52789670ea/epilepsy12/kpi.py#L94)

```python
england_row = create_kpi_report_row(
    "england", measure_title, kpi, england_kpi_aggregation
)

england_row["Country"] = "England"

wales_row = create_kpi_report_row(
    "wales", measure_title, kpi, wales_kpi_aggregation
)

wales_row["Country"] = "Wales"
```

dc2007git:
My mistake thanks for the suggestion @mbarton. Will push the changes to this PR

mbarton:
Thanks @dc2007git! It's not a mistake on your part, just a suggestion on mine :)

dc2007git:
@mbarton the most recent changes I've pushed to this PR can be discarded, as well as with this entire PR, because this is solved in a more holistic way in PR #926 , which means we can remove the unnecessary columns, reorganise the columns, and add the uk column all in one fell swoop!

Issue #907, opened 2024-04-25
QE King's Lynn


eatyourpeas:
There seems to be an error with King's Lynn specifically where aggregations are counting all children in the organisation, at ICB/NHS region and national levels, but not at Trust level. It looks like they are being filtered out.

mbarton:
[posting summary from a call with Simon, Danny and Michael 29/04/24]

We think the issue is probably that now we have added additional filtering, no cases from QE Kings Lynn should be included in the aggregations. However an organisation aggregation row was added previously and our code at the moment doesn't reset it again to zero, it simply skips over any abstraction level that has no corresponding cases.

@eatyourpeas will try changing the code to always write at all abstraction levels every time. We believe that to be safe as even if we have bugs in our aggregation code it can always be derived from the underlying data - so it's fine to overwrite a previous aggregation row even if there's no corresponding data for the organisation now our additional filtering applies

mbarton:
@dc2007git and I looked at this today. We removed all KPI metrics so they would be recalculated the next time aggregations were run.

```python
BaseKPIMetrics.objects.all().delete()
```

This removed any outdated rows but unfortunately it does not seem to have fixed this issue.

The *Trust* column that was blank before is now missing entirely. The organisation shows zero cases at the top level but then does show cases in the rows below. It also shows a "Local health board" column which should not apply here as we're not in Wales.

mbarton:
This is fixed, we’ll track everything else under other issues

PR #929, opened 2024-05-01
Log aggregation rows that have data but are not updated

mbarton:
Our theory for #907 is that previous versions of the code, applying less rigorous filtering for cases, added aggregation rows for which there is now no longer data with the more restrictive filtering.

Our aggregation code never removes or zeros out these orphan rows, leaving data in the KPI report that is incorrect.

This PR adds some extra logging to identify these rows. We can then manually delete or correct using the Django admin in order to produce the KPI report for #895.

The original commit from @eatyourpeas (25b23dec84fa8f2f84ff8b0b5917685ae8366b6e) set these rows to zero. I think it's better off to start by identifying them so that we can try to preserve at the aggregation level a distinction between organisations who have data but it is filtered out for aggregations and those organisations who have never entered data.


mbarton:
@dc2007git and I have looked at this in person and are happy to merge

PR #933, opened 2024-05-02
Eatyourpeas/aggregation-logs-text-fixes

### LINKED ISSUE #907: QE King's Lynn

eatyourpeas:
### Overview

Another incremental improvement in the aggregation suite.
This ensures logs for any updates or creates to the KPIAggregation models, with a summary of how many models changed, and how many records updated with 0 if unscored.

There are also some fixes to the tests:
1. update the filter to accept only completed cases who have completed a full year of care
2. deprecate the get_filtered_cases_queryset_for which is not used for the aggregations anyway in favour of the newer `filter_completed_cases_at_one_year_by_abstraction_level` so testing closer to real life

### Code changes

1. add `filter_completed_cases_at_one_year_by_abstraction_level` to `__init__.py` to allow easier importing
2. remove the `return` statement after testing for None, and nest any code below this test into a conditional.
3. Change `_register_kpi_scored_cases` function in the e12case factory to include cases who have completed a full year of care
4. deprecate `get_filtered_cases_queryset_for` which was now failing update KPIAggregation table tests, and replace with `filter_completed_cases_at_one_year_by_abstraction_level` which is used in the actual calculations anyway

This needs doing - mostly to make explicit that the aggregation tables now only contain fully scored cases that have completed a full year of care

### Related Issues

Closes #907

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -237,18 +237,24 @@ def calculate_kpi_value_counts_queryset(
         aggregate_queries.update(
             {
                 f"{kpi_name}_passed": Count(
-                    DJANGO_CASE(When(**{f"{kpi_name}": 1}, then=1))
+                    DJANGO_CASE(
+                        When(**{f"{kpi_name}": 1}, then=1),
+                    ),
                 ),
                 f"{kpi_name}_total_eligible": Count(
                     DJANGO_CASE(
-                        When(Q(**{f"{kpi_name}": 1}) | Q(**{f"{kpi_name}": 0}), then=1)
+                        When(Q(**{f"{kpi_name}": 1}) | Q(**{f"{kpi_name}": 0}), then=1),
                     )
                 ),
                 f"{kpi_name}_ineligible": Count(
-                    DJANGO_CASE(When(**{f"{kpi_name}": 2}, then=1))
+                    DJANGO_CASE(
+                        When(**{f"{kpi_name}": 2}, then=1),
+                    ),
                 ),
                 f"{kpi_name}_incomplete": Count(
-                    DJANGO_CASE(When(**{f"{kpi_name}": None}, then=1))
+                    DJANGO_CASE(
+                        When(**{f"{kpi_name}": None}, then=1),
+                    ),
                 ),
             }
         )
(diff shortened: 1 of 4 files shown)


epilepsy12/common_view_functions/aggregate_by.py
```diff
@@ -345,80 +345,173 @@ def update_kpi_aggregation_model(
...
+                    logger.debug(f"created {new_obj}")
+                else:
+                    logger.debug(f"updated {new_obj}")
             
-    
-    logger.info(f"{len(list_of_updated_abstraction_level_instance)} scored {abstraction_level.name} instances updated with aggregated scores of a total {AbstractionKPIAggregationModel.objects.filter(cohort=cohort).count()} {abstraction_level.name}s")
+    # value_counts for this abstraction level  already updated
+    # Set all measures to 0 for remaining abstraction levels
+    logger.debug(f"{len(list_of_updated_abstraction_level_instance)} scored {abstraction_level.name} instances updated with aggregated scores of a total {AbstractionKPIAggregationModel.objects.filter(cohort=cohort).count()} {abstraction_level.name}s")
+    empty_kpis = {
```
mbarton, line 419: I think we can probably avoid writing the zero rows now we know that we must clear out all the aggregations rows and re-seed if we change the case filtering criteria at the top of the pipeline

epilepsy12/common_view_functions/aggregate_by.py
```diff
@@ -265,10 +265,10 @@ def calculate_kpi_value_counts_queryset(
         kpi_value_counts = KPI.objects.filter(
             registration__id__in=filtered_cases.values_list("registration")
         ).aggregate(**aggregate_queries)
-    else:
+    else: 
         kpi_value_counts = (
             KPI.objects.filter(
-                registration__id__in=filtered_cases.values_list("registration")
+                registration__id__in=filtered_cases.values("registration")
```
mbarton, line 271: Should this still be `values_list` as on line 266? I think if we change one we should change both

epilepsy12/common_view_functions/aggregate_by.py
```diff
@@ -345,80 +345,173 @@ def update_kpi_aggregation_model(
     
     # update models where numbers have changed.
     for value_count in kpi_value_counts:
-        ABSTRACTION_CODE = value_count.pop(f"organisation__{abstraction_level.value}")
-        if ABSTRACTION_CODE is None:
-            # we don't have any values for this abstraction level (eg local health board only applies in Wales not England)
-            return
+        ABSTRACTION_CODE = value_count.pop(f"organisation__{abstraction_level.value}") #value_count.get(next(iter(value_count))) # 
+        if ABSTRACTION_CODE is not None:
+            # the last value in each abstraction_level is None - discard
```
mbarton, line 350: I agree with removing the early return but maybe the comment is clearer above the if?

I also don't think the last value is None, I think it's if we're grouping by an abstraction level not relevant here eg by local health board in England

epilepsy12/tests/common_view_functions_tests/aggregate_by_tests/test_update_kpiaggregation_model.py
```diff
@@ -174,15 +175,31 @@ def test_update_kpi_aggregation_model_all_levels(
                 **{abstraction_relation_instance_key: abstraction_relation_code}
             ).first()
 
-            kpi_aggregation_model_instance = (
-                abstraction_kpi_aggregation_model.objects.get(
-                    abstraction_relation=abstraction_relation_instance,
-                    cohort=6
-                )
-            )
+            # kpi_aggregation_model_instance = (
```
mbarton, line 178: remove?

mbarton (commented):
Looks fine, just some small comments

mbarton:
@eatyourpeas what's left to get this merged? :)

eatyourpeas (commented):
I am not sure how much extra this PR really adds any more. The new filter function is already in live, so this was really about the logs, the zero rows and maybe a test update.
Perhaps this is one we can do together or we could just close if it nolonger adds value?

eatyourpeas:
I think park this PR for now

PR #941, opened 2024-05-02
Fix KPI report rows

mbarton:
Previously we were looking up the rows in the KPI aggregation models by index as we enumerated the related constants. This meant we looked up the wrong row.

Now we simply iterate through the rows themselves and ask Django to do a join to get us the key field we need from the related abstraction level (eg Trust, ICB).


mbarton:
we mob coded this change so we're happy to merge once CI passes

PR #944, opened 2024-05-07
Refactor to use consistent helper functions when generating the KPI report

mbarton:
When @eatyourpeas, @dc2007git and I mob coded on https://github.com/rcpch/rcpch-audit-engine/pull/941 we felt like there was a simpler way to express the code but focused on fixing the report with the minimum possible changes.

This PR tries to capture as many simplifications as possible before the context inevitably falls out of my head!

The core change is to use the functions from aggregate_by.py consistently across all sheets we generate in the report. To achieve this I broke them out into two:

- `get_kpi_aggregation_rows`
  - Extracts all the rows from the KPI aggregation model and adds an additional `key_field` (eg Country, NHS region...)
- `create_kpi_aggregation_dataframe`
  - Now just loops through all the measures which a4rre hard-coded at the top of kpi.py as not all measures are included in the report
  - It then uses the existing `create_kpi_report_row` function to add the row to the dataframe. That function defends against missing/null/zero field values

These changes allow us to push the sheet specific logic back up into `kpi.py`, for example combining local health boards in Wales and trusts in England when generating the "HBT" sheet.

I've also linked the definition of the measures directly to the fields on the models. This doesn't change any functionality but makes it clearer they're linked and in the unlikely event we change them we would automatically update here.

That means we can pass the Django field definition all the way down to `create_kpi_report_row`, meaning one less parameter for all the function calls. This is maybe not clearer than before but felt like a logical follow on from linking directly to the model definitions at the top level of `kpi.py`.

**Questions**

- [x] ~Our current code and the [template spreadsheet](https://github.com/rcpch/rcpch-audit-engine/issues/791#issue-2128940055) group all the rows by measure except for the HBT sheet where it's grouped by ODS code. This refactor has broken that. Do we need to maintain that behaviour?~
  - Confirmed with @nikyraja the order doesn't matter

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -870,19 +870,18 @@ def create_kpi_report_row(key, kpi_field, aggregation_row, level):
 
 def get_kpi_aggregation_rows(
     model_aggregation,
+    cohort,
     abstraction_key_field,
-    cohort
 ):
     return model_aggregation.objects.filter(
         cohort=cohort,
         open_access=False
     ).annotate(
         key_field=F(f"abstraction_relation__{abstraction_key_field}")
-    )
+    ).values()
 
 def create_KPI_aggregation_dataframe(
     aggregation_rows,
-    cohort,
     measures,
     title,
 ):
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -883,7 +883,9 @@ def get_kpi_aggregation_rows(
             key_field=F(f"abstraction_relation__{abstraction_key_field}")
         )
     
-    # Eagerly evaluate the query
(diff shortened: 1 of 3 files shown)


epilepsy12/common_view_functions/aggregate_by.py
```diff
@@ -838,7 +838,10 @@ def ___delete_and_recreate_all_kpi_aggregation_models():
 """
 Functions to create Excel reports
 """
-def create_kpi_report_row(key, measure, kpi, aggregation_row, level):
+def create_kpi_report_row(key, kpi_field, aggregation_row, level):
+    kpi = kpi_field.name
```
mbarton, line 842: I've passed the Django field definition all the way down here which means one less parameter for the function calls. I'm not sure if this is clearer than passing two parameters as we did before and pulling them out of the model at the top level call in `kpy.py`. Happy to change back depending on how we feel.

Issue #942, opened 2024-05-03
view more patients at once


AmaniKrayemRCPCH:
request from user
- they'd like to see more patients at once (e.g. view 50 patients in one page).

eatyourpeas:
closed in #951

Issue #937, opened 2024-05-02
[Feedback from E12 Webinar 2024.05.02] "Please can the message box not pop / be displayed over the date of first assessment."


pacharanero:
> Please can the message box not pop / be displayed over the date of first assessment.

eatyourpeas:
closed in #951

Issue #906, opened 2024-04-25
add cause - 2q23.3 microdeletion syndrome


AmaniKrayemRCPCH:
2q23.3 Microdeletion Syndrome

There's no associated SNOMED code. 
Approved by Colin.

eatyourpeas:
closed in #951 - live and staging updated

Issue #884, opened 2024-04-22
Case ID column in cohort view


AmaniKrayemRCPCH:
Thanks for adding the case ID on hover. 
Users aren't realising that the case IDs are there and are still emailing with patient identifiable information. Can this be added as a column in the table.

If space is tight, we can remove the 'days remaining' column.

eatyourpeas:
closed by #951

PR #951, opened 2024-05-09
only apply filtered_case_list if not none in case_list template

### LINKED ISSUE #943: pagination case list error
There is a pagination error raised if a filter is applied to the case table and then one of the pagination buttons is selected

eatyourpeas:
Fixes #943

### Overview

Current behaviour - if a sort has been applied to a column in the the case table, the pagination buttons then fail as the filtered_case_list parameter is None. 
What should happen:
if a sort has been applied to a column, the sort flag is passed to the htmx get request, but not the filtered_case_list, allowing pagination to proceed

### Code changes

A conditional is applied in the case_list template partial to test for when filtered_case_list is None, and change the htmx get url params accordingly

### Related Issues

Fixes #943


eatyourpeas (commented):
Additional commit added as related issue and fix to close #942
User request to increase case table size to accommodate 50 children

eatyourpeas (commented):
Additional commit to fix and close #937 - this changes the sweetalert2 config to reposition the modal to top middle of the screen, so not obscuring the date of first paediatric assessment

eatyourpeas (commented):
Additional commit adds 2q23.3 Microdeletion Syndrome to migration__006
closes #906

eatyourpeas (commented):
Additional commit closes #884 and #887
This adds the E12 ID as an extra column to the case table and moves the info button into that column, viewable now only to rcpch audit team, and shows the lead site name. This add a new template filter tag to allow this to be possible

PR #960, opened 2024-05-13
Bump Gunicorn to address dependabot alert

mbarton:
https://github.com/rcpch/rcpch-audit-engine/security/dependabot/3

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -31,7 +31,7 @@ plotly==5.19.0
 nhs-number==1.3.4
 
 # live application server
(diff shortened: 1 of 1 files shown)


eatyourpeas (approved):
I am not sure how purist to be about this? I have rebuilt a fresh instance locally and it all works with the new gunicorn, but the only way to test it I guess is in production? Or perhaps in staging?
Anyway, approving here and leav to you to merge?

mbarton:
Agreed lets merge and revert in the unlikely event it causes any issues in deployment

Issue #971, opened 2024-05-16
Orphan KPI and AuditProgress records


eatyourpeas:
When a Case is deleted, the registration and associated records are deleted, but not the KPI and AuditProgress records also.

To fix this issue will require two steps:
1. refactor the delete function in `case_views.py`
2. In the django shell in live/staging/development, remove all orphan KPI and AuditProgress records which have no registration instance

eatyourpeas:
when #972 merges that will fix this issue, but leaving open here as still need to prune the databases of orphan records

eatyourpeas:
Now merged. There are 278 orphan/empty KPI/AuditProgress records in live to remove

eatyourpeas:
These have now been removed. More are likely to appear until these changes in development are merged in to live, but hopefully we can tidy again at that point.

Issue #893, opened 2024-04-22
Viewing locked audit records


AmaniKrayemRCPCH:
If a user is accidentally assigned to cohort 5, I can't easily go in and edit that
Similarly, I can't view 'submitted' records without unsubmitting them. 

Is it possible to do the following:
* for records where the deadline has passed, allow users to open and read these. Allow RCPCH staff to edit these
* for submitted records, allow all users to open and read these. These still need to be 'unsubmitted' for anyone to make edits.

Let me know if there's a better solution.

eatyourpeas:
This might be hard to implement @AmaniKrayemRCPCH because by re-enabling the row, the users existing permissions would apply: it is of course possible to give them editing permission for one cohort, and view only permission for another, but will take a bit of time. I could potentially give just RCPCH access to o those disabled rows - that would be the quickest fix to this. This way, all cohorts would be viewable and editable to RCPCH only - could that work?

AmaniKrayemRCPCH:
agreed that it would be easier to give RCPCH staff edit access - as long as we're able to change the FPA date of a patient in cohort 5, or to delete them.

eatyourpeas:
Just to add, this vulnerability should be historical. As per this code snippet below - the first paediatric assessment date minimum value now is set at the currently submitting cohort (currently 6), but earlier (date of birth) for RCPCH audit team
```python
if request.user.is_superuser or request.user.is_rcpch_audit_team_member:
        earliest_allowable_date = case.date_of_birth
    else:
        # registering a new child in the audit by a clinical team
        # sets the minimum allowable date to the currently submitting cohort start date
        earliest_allowable_date = cohorts_and_dates(
            first_paediatric_assessment_date=date.today()
        )["submitting_cohort_start_date"]
```

PR #972, opened 2024-05-16
Eatyourpeas/all-rcpch-access-to-disabled-records

eatyourpeas:
### Overview

The E12 would like to be able to edit records which are disabled as are in a historical cohort (eg 5). This used to happen in error when users put in a first paediatric assessment date that was wrong, but now the minimum date for users to select is the start of the currently submitting cohort.

In the course of investigating this it was discovered that the Django admin Registration, KPI and AuditProgress would not render if there were orphan records. This PR therefore also fixes:
1. cascade case deletions to include AuditProgress and KPI related records
2. Admin template rendering if orphan records present

### Code changes

Update to documentation to clarify first paediatric assessment date minimum values for users

Fix AuditProgress and KPI and Registration models to return sensible string even if orphan records

Fix `case_views.py` to cascade deletes to to KPI and AuditProgress related records

Update `case_table` partial to allow RCPCH users access to disabled records

Please describe any changes to documentation here.

### Related Issues

List any issues related to this PR here.

### Mentions

### DIFF
--- a/documentation/docs/development/date-validations.md
+++ b/documentation/docs/development/date-validations.md
@@ -80,7 +80,7 @@ It raises a ValueError which is caught in the UI.
 
 | Model | Date | mandatory | earliest allowable date | other flags |
 | ---- | ---- | ---- | ---- | ---- |
-| Registration | first_paediatric_assessment_date | Yes | date_of_birth |   |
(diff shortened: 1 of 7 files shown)


eatyourpeas (commented):
this will close #971 and #893

PR #1003, opened 2024-06-24
Hosted Azure Database

mbarton:
Support entra ID authenticated hosted azure flexible postgres


mbarton:
Merging as the migration is already in progress and we need this code on live. If there are any CI failures we will address them with follow up PRs

PR #1004, opened 2024-06-27
Refresh Azure Postgres access token

mbarton:
Adds a manage.py command to get a new Azure Postgres access token and write it to a Postgres [connection service file](https://www.postgresql.org/docs/current/libpq-pgservice.html).

By triggering the command on a cron we can ensure that the access token is updated before expiring. This fixes an issue seen in production where Django tried to establish a new connection but the access token had expired.


epilepsy12/management/commands/write_azure_pg_service_file.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Command(BaseCommand):
+    help = "Generate a Postgres service file to use an Azure managed identity to authenticate with the database."
+    
+    def fetch_azure_token(self):
+        url = 'http://169.254.169.254/metadata/identity/oauth2/token'
+
+        params = {
+            'api-version': '2018-02-01',
+            'resource': 'https://ossrdbms-aad.database.windows.net'
```
mbarton: It turns out the `client_id` parameter was optional as we only have a single managed identity attached to the VMs

mbarton:
This doesn't work - tested in staging and it complains the password line is too long in the service file

```
django-1   |   File "/usr/local/lib/python3.11/site-packages/django/db/backends/postgresql/base.py", line 277, in get_new_connection                          
django-1   |     connection = self.Database.connect(**conn_params)                                                                                            
django-1   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                            
django-1   |   File "/usr/local/lib/python3.11/site-packages/psycopg2/__init__.py", line 122, in connect                                                      
django-1   |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)                                                                       
django-1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                       
django-1   | django.db.utils.OperationalError: line 4 too long in service file "/root/.pg_service.conf"
```

I will try and put it in a separate [Postgres password file](https://www.postgresql.org/docs/current/libpq-pgpass.html) instead

mbarton:
Tested this again with the password file and it's working :)

Once merged we'll need to edit the crontab on each VM manually to trigger `python manage.py write_azure_pg_password_file` on a schedule

PR #1019, opened 2024-08-11
make-description-optional-*sad face*

### LINKED ISSUE #1015: Make seizure description question non mandatory
Some users have fed back that they would like the free text box within the episode description questions (4 Multiaxial diagnosis, DESSCRIBE definitions) to be made non-mandatory.
Seziure description screenshot
- The Yes/No question would remain mandatory, but users would have the choice to add a free text description

The Methodology & Dataset group agreed with this change. 

Would this be possible? Would it mess too much with the scoring of question completion etc?

eatyourpeas:
### Overview

The E12 data group have accepted a request from some users requesting the episode free text description to be made non-mandatory. The description field uses fuzzy logic to match any key words (taken from a validated list produced by the american international league against epilepsy ILAE group). The original reason to include the feature in the early build had been to do frequency analysis on terms in the free text and improve on this slightly odd and eclectic American list of terms (which includes words like 'fencers posture' but does not include words like 'shaking'). By making this an optional field, it seems unlikely now it will be completed by many users hence the *sad face*, and depending on how much it is completed, the validity of any data collected will be watered down. Sorry this is only an audit, I should possibly not be quite so involved...

Anyway, this PR removes the `description` field from the scoreable fields and amends the signposting to users to reflect this in the templates. It also updates the tests which check the form scoring.

### Code changes

- `form_calculation_constants`: this file stores the minimum scoreable fields. `description` is commented out with a reference to this issue for future.
- `recalculate_form_generate_response`: omits `description from the scored fields
- test files: these remove description from the totals
- `description.html` and `description_labels.html`: update the tooltip and message to signpost to users that the field is optional.

### Related Issues

closes #1015

### DIFF
--- a/epilepsy12/common_view_functions/recalculate_form_generate_response.py
+++ b/epilepsy12/common_view_functions/recalculate_form_generate_response.py
@@ -152,7 +152,6 @@ def completed_fields(model_instance):
                                 counter += 1
                     else:
                         counter += 1
-        print(f"field: {field.name} - counter: {counter}")
     return counter
 
 
--- a/epilepsy12/common_view_functions/recalculate_form_generate_response.py
+++ b/epilepsy12/common_view_functions/recalculate_form_generate_response.py
@@ -111,7 +111,7 @@ def completed_fields(model_instance):
             if getattr(model_instance, field.name, ()) is not None:
                 if (
                     field.name == "epilepsy_cause_categories"
-                    or field.name == "description"
+                    # or field.name == "description"
                     or field.name == "mental_health_issues"
                 ):
                     if len(getattr(model_instance, field.name)) > 0:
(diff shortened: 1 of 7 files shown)


epilepsy12/common_view_functions/recalculate_form_generate_response.py
```diff
@@ -519,7 +520,7 @@ def expected_score_for_single_episode(episode):
     # essential fields already accounted for are:
     # seizure_onset_date
     # seizure_onset_date_confidence
-    # episode_definition
+    # episode_definition - DEPRECATED
```
mbarton, line 523: Reading the thread in #1015 I think it's more optional than deprecated?

mbarton (approved):
Looks good. I've pushed a commit to this PR to update the wording as requested in https://github.com/rcpch/rcpch-audit-engine/issues/1015#issuecomment-2283488021

epilepsy12/common_view_functions/recalculate_form_generate_response.py
eatyourpeas: yes good point.

PR #1023, opened 2024-08-21
Update first_paediatric_assessment_date tooltip

mbarton:
Updated tooltip on first paediatric assessment date to describe what can be included as the initial assessment

### DIFF
--- a/epilepsy12/models_folder/registration.py
+++ b/epilepsy12/models_folder/registration.py
@@ -34,7 +34,7 @@ class Registration(
     first_paediatric_assessment_date = models.DateField(
         help_text={
             "label": "First paediatric assessment",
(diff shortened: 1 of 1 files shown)


mbarton:
I think at least the failure to add an issue is because this PR came from a fork - not sure about the second failure though so re-running to see.

mbarton:
Ah it's exit code 137 which I think means we've blown the 16GB memory limit granted to us by GitHub actions (https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

mbarton:
ok it passes in https://github.com/rcpch/rcpch-audit-engine/pull/1023, not really sure why but I'll merge that one instead

PR #1040, opened 2024-09-07
Allow password reset even if user has never confirmed their email

### LINKED ISSUE #1039: did not click on sign up link
If a user never activates their account they cannot reset their password

eatyourpeas:
### Overview

A common situation is that users fail to click on their account creation link within 72 hours.
*What should happen:*
User contacts Epilepsy12 who navigate to the user in the platform -> edit user -> click on reset password
This triggers a password reset email with a token to the user who clicks through to set a password
This pathway has been retested and works.
*What actually happens:*
User does not contact Epilepsy12 team as directed as does not read the email. User navigates to password reset page and requests new password. Because the user does not yet have a useable password, no email is sent - this is django stock behaviour. There is no messaging to the user to say this, so frustration follows.

Proposed solution offered in this PR
- user account created by their centre lead clinician. An email is sent out with updated details on steps to take to confirm their email in future if they do not click on the link within 72hrs
- user fails to confirm account within 72 hr and enters the gulag
Two options then result:
**Option One**
1. user contacts epilepsy12 who navigate to user account in the platform - this is the original expected workflow. E12 admin click on the resend email button which only appears for users who have not yet confirmed their accounts. The title of this button has been clarified to `Resend Account Confirmation Email`.
2. User gets email with fresh token and has a further 72 hrs to confirm account and set password. User has escaped the gulag.
**Option Two**
1. User does not read any of the emails and navigates to password reset, or reads the original email with the expired token, and follows link to reset password
2. User receives email with reset password link and confirms email as well as resetting password. User has escaped the gulag.

Along with this small change in workflow, better signposting in the user table has been added, where a blue target appears next to the user if they have an active account (all accounts are active by default) but have yet to confirm their email. This is replaced by a pink tick if email has been confirmed. The popups have also been changed to clarify the users status.

## Implications
Currently the email confirmation step is controlled by E12. This new work flow gives control for the email confirmation step to the user. 

The alternative solution would be to retain the current work flow but institute better messaging in the emails and on the login and password reset screen directing users who have not yet confirmed their accounts to contact E12 and ask for the account confirmation email to be resent. 

The only implications of this new proposed workflow are possibly that is a little less secure. In either situation, the only way an account could be created by a bad actor would be for them to intercept the emails or have access to the user's email account. The current workflow allows an E12 audit team member to have communication with the user wishing to confirm their account as an extra layer of security that they are happy the person is who they say they are. In the new work flow, a user that has access to an email address in the platform can reset the password and confirm the account in one step.

### Code changes

The standard django ResetPasswordForm has been subclassed and the `get_users` method has been overridden to return all users with a matching email, so long as they are active, whether they have useable passwords or not. This class is used by the reset password view.

Some wording changes to the emails sent have been made, and to the create user form.

### Related Issues

Closes #1039

### DIFF
--- a/epilepsy12/scratchpad.md
+++ b/epilepsy12/scratchpad.md
@@ -9,4 +9,25 @@ This file is to write steps in a workflow. It is not a substitute for the docs.
 3. check email_confirmed flag in admin as superuser
 4. allow 2 minutes to expire
 5. click on confirm email link as logged out user to confirm token has expired
-6. Try and reset password as new user
+   Option 1
+
+- superuser clicks on resend email link and logs out
+- user clicks on password reset email from this and resets password then logs in
+- superuser checks admin
+
+Option 2
+
+- user has not confirmed account and that token has now expired
+- user tries to reset password - they are in the gulag: no email is sent
+
+1.  Try and reset password as new user
+
+
+Items to fix
+
+
+
+Observations
+The newly created user with email_confirmed being false and token expired has is_active still true, password_last_set datetime is same as creation datetime
+Option 1 workflow is working as expected: user has 
+
--- a/epilepsy12/scratchpad.md
+++ b/epilepsy12/scratchpad.md
@@ -1,33 +1,3 @@
 # Scratchpad
 
-This file is to write steps in a workflow. It is not a substitute for the docs.
-
-## If a user fails to click on the confirm email link they cannot subsequently reset their password. Debug work flow
-
-1. Set token expired to 2 minutes
-2. create new user
-3. check email_confirmed flag in admin as superuser
-4. allow 2 minutes to expire
-5. click on confirm email link as logged out user to confirm token has expired
-   Option 1
-
-- superuser clicks on resend email link and logs out
-- user clicks on password reset email from this and resets password then logs in
-- superuser checks admin
-
-Option 2
-
-- user has not confirmed account and that token has now expired
-- user tries to reset password - they are in the gulag: no email is sent
-
-1.  Try and reset password as new user
-
-
-Items to fix
-
-
-
-Observations
-The newly created user with email_confirmed being false and token expired has is_active still true, password_last_set datetime is same as creation datetime
-Option 1 workflow is working as expected: user has 
-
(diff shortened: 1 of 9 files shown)


epilepsy12/scratchpad.md
```diff
@@ -0,0 +1,3 @@
+# Scratchpad
+
+This file is to write steps in a workflow. It is not a substitute for the docs. It is included for convenience
```
mbarton: Do we need this checked in?

epilepsy12/views/user_management_views.py
```diff
@@ -598,11 +600,12 @@ def form_valid(self, form):
 
 
 class ResetPasswordConfirmView(PasswordResetConfirmView):
-    # overridden custom django password reset work flow. User has submitted a valide password for reset
+    # overridden custom django password reset work flow. User has submitted a valid password for reset
     # their email is stored in session for use later to update logs if they are successful
     def form_valid(self, form):
         # Store the user's email in the session
         self.request.session["user_email"] = form.user.email
+        print(f"Password reset requested for {form.user.email}")
```
mbarton: Use a logger just for consistency?

templates/registration/password_reset_confirm.html
```diff
@@ -49,7 +49,7 @@ <h5>Password Guidance</h5>
         
     {% else %}
         <h1>Password reset failed</h1>
-        <p>The password reset link was invalid, possibly because it has already been used. Please request a new password reset.</p>
+        <p>The password reset link was invalid, possibly because it has already been used or it has expired - tokens only last 72 hours. Please request a new password reset.</p>
```
mbarton: I wouldn't say token here, it doesn't mean anything to a user. Maybe just:

```suggestion
        <p>The password reset link was invalid, possibly because it has already been used or it has expired after 72 hours. Please request a new password reset.</p>
```

mbarton (commented):
I'm not sure this weakens security, email confirmation is more to check the email address is correct in the first place.

Following on from this though should we have an email confirmation step at all?

mbarton:
@SAbdin93 and I have confirmed on a teams call this works - the user was able to register and set 2fa

PR #1029, opened 2024-08-29
Migrate to Azure Container Apps

mbarton:
Sets up E12 to continuously deploy to Azure Container Apps on merge to live

PR #1042, opened 2024-09-11
Fix the docs

mbarton:
When we migrated to Azure Container Apps (#1029) I forgot that the documentation was hosted via `mkdocs serve` and a special route in the Caddyfile. There was nothing in the new setup that reproduced this, so we fell back to the old iframe docs route which referenced a static site that no longer exists.

To put the docs back to how they were I've configured Whitenoise to serve them. Unfortunately this means we need to run `docker compose build` twice because we use the Django container to run mkdocs as we have plugins installed. In practice though all the base layers are already built so it's not much slower.

I've tested this by manually editing the GitHub deploy workflow to deploy to staging from this branch.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - live
-      - mbarton/fix-the-docs
 
 permissions:
   id-token: write
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - live
+      - mbarton/fix-the-docs
 
 permissions:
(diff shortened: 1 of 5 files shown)


mbarton:
Overdue on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 30 minutes and 3 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 34 minutes and 16 seconds ago) Please check your changes!

SAbdin93:
Thanks Michael! :)

PR #1044, opened 2024-09-25
MVP Organisational Audit

mbarton:
Adds a model for the organisational audit and a UI to fill it in.

Screenshot from 2024-09-25 16-28-03

I've implemented the UI as one big form and used HTMX to automatically save changes as you go along. As such the model allows `null=True` and `blank=True` for all fields.

The UI is driven from fields on the model help text.

To test:

- Go to the Django admin
- Create an OrganisationalAuditSubmissionPeriod
- Set it to active
- Visit /trust/1/audit

TODO in this PR:
  - Check we're happy with how the model is defined
  - Finish off the UI for section 6 onwards
    - [x] ~~Support 6.3i which as "other" as an option of a child (make `parent_question_number` work for children too)~~
    - [x] Other support for multiselect (6.4)
    - [x] Don't show automatically generated question numbers for children (especially Other fields)
    - [x] Fix bug where 7.4.2 Other shows if you select Other but then change 7.4 to Yes (ie recursively check visibility of parents)
    - [x] Should 7.2 be a multiselect?
    - [x] Would section 4 be clearer as a multiselect?
  - ~~See if it’s better to replace the recursive lookup helpers with recursive rendering code~~
  - ~~Collapse the migrations to a single file~~

TODO in future PRs:
  - ~~Other text box appears in 5.3 before an option is selected~~
  -  Add a page title
  - Fix focus jumping to start of input after save
  - Render Yes/No/Uncertain inline in section 4
  - Show progress tracking (i.e have you finished the form)
  - Test form validation and show errors
  - Show answers from previous submission for questions that aren't filled in yet
    - The user should proabably "confirm" these answers before we say the form is filled in
  - Download the results as CSV for the RCPCH audit team
  - Remove the other answers if 'Cannot achieve any of the above' selected for 7.4


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 12 minutes and 43 seconds ago) Please check your changes!

PR #1076, opened 2024-10-15
Banners!

### LINKED ISSUE #882: Announcements/Notifications
We want to reduce the amount of comms we're sending via email. Is there a way we can share comms via the platform e.g. a notification icon at the top right, or alerts/pop-ups when users log in.

These will typically be used to announce things like - upcoming deadlines, publications, events & webinars

### LINKED ISSUE #1059: Add organisational audit banner to organisation dashboard
Add a banner onto the organisation dashboard, visible to Lead clinicians and us, to direct people to the organisational audit when open.

“The 2024 Organisational Audit is now open. Please complete data entry[link] by 14 January 2025.”
- Either hyperlink or via additional button.

### LINKED ISSUE #1060: [Organisational audit] - description at top of submission page
Add the following description at the top of the organisational audit page:

“Please complete the 2024 Organisational Audit for your Trust/Health Board, to describe your epilepsy service as of 1 November 2024. Please complete your data entry by 23:59 on 14 January 2025. 
This form has been pre-populated with data from your most recent organisational audit submission. Please note, you will still need to submit/accept the data before the above deadline.”

Have submit/accept button either at top or bottom of the page – only visible if the form is prepopulated.

mbarton:
Fixes #1060
Fixes #1059 
Fixes #882 

Add banners using the Django admin, targeted to specific pages using a regular expression. Add arbitrary HTML to the banner.

Screenshot 2024-10-15 13 10 13

Screenshot 2024-10-15 12 42 18

Allows us to implement the guidance for the organisational audit in #1060 and #1059 generically. In particular you can target banners at specific users (eg the organisational audit CTA should only be shown to lead clinicians). RCPCH audit team members can always see banners.

Banners are a model in the database but are cached for 10 seconds to avoid introducing another database call on every page load. This might be overkilll though as our pages load from the DB pretty much every time anyway. Although this way at least the unauthenticated homepage doesn't hit the DB every time.

@eatyourpeas do you think a model and the Django admin is the right way to implement this? I'm wondering should the banner definitions just be in the codebase since we have a quick CD pipeline now? At least having it in the DB the team can change banners without raising a PR and being worried they'd break something.


eatyourpeas (commented):
This is a really nice feature.
I have noticed that page load is slow. If you change organisations for example there is quite a lag. I am a bit surprised by this because I think we should not be doing any aggregations on page load, only pulling the table from the db. When we took celery out though I am not sure now how we left that. Might be worth seeing how much work the db is doing and what time it is taking: otherwise you are right it might add to latency.
I like the idea of using the db though - we could use admin to start with but It would not be much work to create a form I guess for RCPCH audit team members to CRUD banners themselves, with a drop down to target different user groups. 
The nice thing about a model is that you could potentially track who has seen the banner and who hasn't? So for example if they have already completed the audit you could hide the banner, or if you have read the notification we could stop showing it. I find it annoying at work for example to continue getting emails telling me to fill out the staff survey if i have already filled it in.

mbarton:
Cool thanks. I agree that in the context of the slow page load the cache doesn't make much difference - testing it locally it takes 1.5 seconds without the banners and 1.6 with.

I think I will still keep in it though just for the sake of the unauthenticated homepage.

I've raised a separate issue to look at the performance of the org summary page: https://github.com/rcpch/rcpch-audit-engine/issues/1092

mbarton:
> The nice thing about a model is that you could potentially track who has seen the banner and who hasn't? So for example if they have already completed the audit you could hide the banner, or if you have read the notification we could stop showing it

I'm definitely up for the latter, the former feels quite use case specific where I'm trying to keep the banners implementation as general as possible

mbarton:
Overdue on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 30 minutes and 5 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 57 minutes and 59 seconds ago) Please check your changes!

PR #1087, opened 2024-10-16
Fix crash for users without an org employer

mbarton:
RCPCH audit team members might not have an organisation employer, but they're still allowed to see all the org audit pages. Fix crash.


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 20 minutes and 38 seconds ago) Please check your changes!

Issue #1096, opened 2024-10-30
Replace findthatpostcode with our own self-hosted postcodes.io


mbarton:
We have noticed variable performance on findthatpostcode.uk, sometimes up to 10 seconds for a single request. This is fine because we are of course not paying for it! However in practice that's likely to be too slow for our users.

We have discussed creating a self-hosted instance of [postcodes.io](https://postcodes.io/). We moved away from that provider as it seemed less up to date than findthatpostcode but since then they have made multiple releases and should now be up to date as of August 2024.

It's now up and running at https://api.rcpch.ac.uk/postcodes (API key required).

We will migrate in two steps

1. Create new settings that when set use our postcodes.io
2. Remove the old findthatpostcode code path

That way we're rollback safe and won't immediately break everyones local dev environments.

We also need to handle terminated postcodes. Postcodes.io provides a separate endpoint for this, we can fallback to that if the main one gives us a 404.

As part of this we should consider porting these postcode handling improvements over from NPDA:

- https://github.com/rcpch/national-paediatric-diabetes-audit/pull/970
- https://github.com/rcpch/national-paediatric-diabetes-audit/pull/973
- https://github.com/rcpch/national-paediatric-diabetes-audit/issues/985

mbarton:
Note the docs need updating when this change is made: https://github.com/rcpch/rcpch-audit-engine/pull/1105/files/3fb2de2aa0a0f48b118dc4c4bf932a2686b15b1d#r1867521080

mbarton:
I've seen a couple of cases over the past 24 hours where correct postcodes are rejected because findthatpostcode returns a 429 Too Many Requests

mbarton:
Bumping the priority of this as I've seen another 429 this morning. It would also "fix" #1173 for future patients by disallowing `O` as the sector when it should be `0`.

Looking at the error message the free allocation is per day, so we might be stuck all day now?

![Image](https://github.com/user-attachments/assets/47efcae8-37c8-4814-9091-fc021b71f252)

mbarton:
postcodes.io does have a terminated postcode search that returns geo coordinates so for retired postcodes we could get the IMD that way (see https://github.com/rcpch/national-paediatric-diabetes-audit/issues/716)

PR #1105, opened 2024-11-24
jersey-joins-epilepsy12

### LINKED ISSUE #1100: Adding Jersey to platform
Jersey will be joining the E12 audit from April 2025. This will require us to make a few changes on the platform, most things you may be aware of from the NPDA:

- [x] Allowing the platform to accept URN numbers for Jersey instead of NHS numbers
- [x] Updating the maps to show Jersey
- [x] Possibly updating their KPI dashboard?

This will be their reporting structure:
•	Organisation: Jersey General Hospital
•	Trust: Jersey General Hospital
•	ICB: N/A
•	OPEN UK: SWIPE (South West Interest Group Paediatric Epilepsy)
•	NHSE Region: N/A
•	Country: Jersey
•	National – not included (so remains as England and Wales only)

The deadline for this is **1 April 2025**.

### LINKED ISSUE #1101: Adding Jersey to Organisational Audit
Related to #1100, Is it possible to add Jersey to the 2024 organisational audit? Can we give them just access to the org audit form, without us needing to have the clinical side worked out?

eatyourpeas:
Fixes #1100
Fixes #1101 

### Overview

Jersey is in the Channel Islands and part of the UK but does not participate in the NHS. There are reciprocal agreements about some hospital treatment, but inpatient care is free only to people who have been resident for 6 months and have a Health Card.

Adding Jersey General Hospital to Epilepsy12 has implications for 7 parts of the audit

- **Organisational structure**: the E12 structure is that organisations have a Trust or Local Health Board as their parent. There is a hierarchy above this which varies between England and Wales. Organisations and Trusts in England might have the same name, but will always have separate ODS codes, which are often similar. Jersey General Hospital in St Helier is a Trust which provides medical care directly. To work around this, Jersey General Hospital has been created both as an Organisation and a Trust, each with the same ODS code, so that it is its own parent.
- **Levels of Abstraction**: Jersey has been added as a separate country, so that it can report at the level of organisation, trust, OPEN UK Network, and country, though the numbers for 3 of these 4 hierarchies will be the same.
- **Boundary files**: Shape files for Jersey's boundaries have been included in the `shape_files` folder. The boundaries do not map evenly to the Country model, so the Country model has been refactored to allow some fields to accept blank values. There are 3 migrations that have been created for this - the first to refactor the boundary/Country model and Trust model (to accept longer ODS codes), the second to map the Jersey shape file layers to the Country model and import the data, the third to add Jersey organisation and Trust to the database if does not exist.
- **Maps**: The functions relating to mapping have been slightly altered to filter to country (something that really needed fixing anyway). The template filters out all English and Welsh organisational geographies.
- **Postcodes**: E12 currently still uses `findmypostcode.co.uk` as a postcode validation and lookup service but RCPCH have more recently moved over to its own instance of `postcodes.io`. Neither of these support Jersey so a new function is included to validate Jersey postcodes (all of which begin JE) with regex. This won't validate the postcode if it does not exist, but will ensure it is at least more or less in the right format. As part of this the seed functions and E12CaseFactory have been updated to generate Jersey postcodes from a random list.
- **GeoCodes**: E12 currently looks up postcodes against an API which returns longitude and latitude, and these are stored in the model (using SRID 27700) and this is used to plot them on the scatter plots. There is no authoritative open source lookup of Jersey postcodes matched with geocodes. The Jersey government have 'private partners' who sell this data at £85 a year as a csv download is the cheapest I can find. Or there is Google maps API or a string of other providers such as OpenCage, but these are similarly paid for services. There is not much about this when I search, but generally people in Jersey seem to say that accuracy and coverage is not that great. For the moment therefore  a hacky work around is not to save geocode data on save and skip the lookups if the postcode is a Jersey one. This means the scatterplots currently will never have plots. I will open a separately to track the progress of this.
- **Index of multiple deprivation quantiles**: We currently get IMD quantiles from our own service which accepts a postcode across England, Wales, Scotland and Northern Ireland and returns country-specific quantiles. It does not accept Jersey postcodes, since the reference data for IMD does not cover Jersey. Jersey released some data based on its census of 2021 [here](https://www.gov.je/SiteCollectionDocuments/Government%20and%20administration/R%20Census%202021%20Indicators%20of%20Deprivation%20Report%2020221213%20SJ.pdf) and makes interesting reading. In the 10 years from 2011, deprivation has fallen, completely the opposite of England who introduced austerity measures in 2010. There are more old people in Jersey, and more people now have cars, fewer have children under 5 y, more have qualifications and their own home. Interestingly more are single parents. Because the population is small (~100,000), IMD is reported in vingtaines, rather than quintiles or deciles. The actual data is not published (at least that I can see) be we could contact Ian Cope (Chief Statistician | chiefstatistician@gov.je) potentially to ask if we may have it. I should have thought a simple look up could be hard coded into E12, though ideally we would include this in the RCPCH Census Platform. For the moment, we get a 500 error (which needs fixing) and this is trapped in a `try..except` so that no value is stored.
- **Unique Reference Number (URN)** Since Jersey is not part of the NHS, users do not have NHS numbers, instead they have URNs. There is no published methodology on the creation of these numbers, how uniqueness is ensured, whether there is any checksum/validation method as is the case for NHS numbers. To accommodate this this PR includes a new field in the `Case` model `unique_reference_number` which is (unique, obviously) but conditionally used in the forms, list views and so on if Jersey General Hospital and its patients are involved. The `CaseForm` now accepts an `organisation_id` parameter to check for this and has a `is_jersey` flag to decide whether to allow users to store URNs or NHS Numbers for a given patient. Cleaning methods have been updated for this. I have not yet added any tests for this but it appears to work in the UI at least. The templates have been updated conditionally to reference URN vs NHS Numbers, as well as the filters.

### Other bits

The admin has been updated to allow search fields for Trust and Organisation models

Documentation has been added on how to add .shp files and laymapping as this process is convoluted and easy to get wrong. I will add some of the detail above in separate commits.

### Tests
There are no tests specifically for Jersey. I will do this in separate commits. Existing tests are all passing.

### Related Issues

This closes #1100

New issues will be created for:
1. service to validate Jersey postcodes
2. service to return longitude/latitude against Jersey postcodes
3. index of multiple deprivation in Jersey


eatyourpeas:
image
image
image

documentation/docs/development/postcodes.md
```diff
@@ -0,0 +1,14 @@
...
+reviewers: Dr Simon Chapman
+---
+
+This should be read in conjunction with sections on mapping and index of multiple deprivation. 
+
+Postcodes for patients are stored in RCPCH-Audit-Engine securely as part of the national agreement for the RCPCH Epilepsy12 clinical audit. Opt out can be requested by patients and all data with the exception of the platform unique identifier (not the NHS number or Unique Reference Number) which is retained to generate an accurate denominator.
+
+Postcodes are used primarily to calculate indices of multiple deprivation, but are also used to provide a scatterplot for clinicians of patients in a given organisation to report maximum/minimum/mean and median distances patients have to travel for care.
+
+Postcodes are passed to an RCPCH hosted service which has its own instance of the open source project maintained at [postcodes.io](https://postcodes.io). This reports information against postcode which include LSOA (see Indices of Multiple Deprivation) as well as longitude and latitude. These latter data points are used for scatter plots.
```
mbarton: This isn't the case yet, it's still on findthatpostcode.uk (https://github.com/rcpch/rcpch-audit-engine/issues/1096)

```suggestion
Postcodes are passed to findthatpostcode.uk. This reports information against postcode which include LSOA (see Indices of Multiple Deprivation) as well as longitude and latitude. These latter data points are used for scatter plots.
```

epilepsy12/common_view_functions/map_from_shape_file.py
```diff
@@ -35,7 +38,17 @@ def generate_case_count_choropleth_map(
     """
     px.set_mapbox_access_token(settings.MAPBOX_API_KEY)
 
-    region_tile = region_tile_for_abstraction_level(abstraction_level=abstraction_level)
+    # if organisation.ods_code == "RGT1W" and (
```
mbarton: Still required?

epilepsy12/forms_folder/case_form.py
```diff
@@ -79,19 +84,68 @@ class CaseForm(forms.ModelForm):
...
+                country_boundary_identifier=Organisation.objects.get(
+                    id=self.organisation_id
+                ).country.boundary_identifier,
+                is_jersey=self.is_jersey,
+            )
+            if self.is_jersey:
+                # this is Jersey
+                if self.instance.unique_reference_number is None:
+                    # this is a new form - create a new URN that is unique
+                    while True:
```
mbarton: It is gated behind a DEBUG flag but I'm always a bit scared at having a `while True` buried within nested code - it's a risk that this could run unexpectedly and crash the app.

Do we need it? We don't bother with it for NHS numbers below. We could also just use a UUID

epilepsy12/forms_folder/case_form.py
```diff
@@ -124,6 +178,11 @@ def clean_date_of_birth(self):
 
     def clean_nhs_number(self):
         # remove spaces
+        organisation = Organisation.objects.get(id=self.organisation_id)
```
mbarton, line 180: What use case does this cover?

epilepsy12/forms_folder/case_form.py
```diff
@@ -141,8 +200,37 @@ def clean_nhs_number(self):
...
+        nhs_number = cleaned_data.get("nhs_number")
+        unique_reference_number = cleaned_data.get("unique_reference_number")
+        if self.is_jersey:
+            if unique_reference_number is None:
+                raise ValidationError("URN is a mandatory field.")
+            if self.instance.unique_reference_number:
+                # there is an existing URN
+                if unique_reference_number != self.instance.unique_reference_number:
+                    # the new URN does not match the one stored
+                    if Case.objects.filter(
```
mbarton: We should implement this as a unique constrant at the database level not as a cleaner

epilepsy12/management/commands/seed.py
```diff
@@ -168,6 +168,9 @@ def run_dummy_cases_seed(cases, organisations, noskip, verbose=True):
         num_cases_to_seed_in_org = int(cases / len(organisations))
         print(f"Creating {num_cases_to_seed_in_org} Cases in {org}")
 
+        # Check if the organisation is in Jersey - this is important for generating postcodes and URNs rather than NHS Numbers
+        is_jersey = org.country.boundary_identifier == "JEY"
```
mbarton, line 173: We should add Jersey to the ORGANISATION_SEED_LIST so we have patients there in our local dev instances to tease out any unexpected bugs

epilepsy12/models_folder/case.py
```diff
@@ -64,7 +64,20 @@ class Case(TimeStampAbstractBaseClass, UserStampAbstractBaseClass, HelpTextMixin
...
     nhs_number = models.CharField(  # the NHS number for England and Wales
-        "NHS Number", unique=True, blank=True, null=True, max_length=10
+        "NHS Number",
+        unique=True,
+        blank=True,
+        null=True,
+        max_length=10,
+        help_text="This is the NHS number for England and Wales. It is used to identify the patient in the audit.",
+    )
+    unique_reference_number = models.CharField(
```
mbarton, line 74: As mentioned above, I think we could use a [UniqueConstraint](https://docs.djangoproject.com/en/5.1/ref/models/constraints/#uniqueconstraint) across NHS number and URN rather than enforcing uniqueness for URNs at the form cleaner level

epilepsy12/views/case_views.py
```diff
@@ -641,13 +661,15 @@ def update_case(request, organisation_id, case_id):
...
     organisation = Organisation.objects.filter(pk=organisation_id).get()
+    case = get_object_or_404(Case, pk=case_id)
+    form = CaseForm(instance=case, organisation_id=organisation_id)
 
     # set select boxes for situations when postcode unknown
-    country_choice = ("ZZ993CZ", "Address unspecified - England")
+    country_choice = (
+        "ZZ993CZ",
+        "Address unspecified - England",
+    )  # TODO we are using the same value for England and Jersey. is that correct?
```
mbarton, line 693: I guess not but I'm not sure it's worth the conditional logic?

@nikyraja @SAbdin93 Should we add a new ZZ code for unspecified addresses in Jersey?

epilepsy12/forms_folder/case_form.py
eatyourpeas: good point. will remove

epilepsy12/forms_folder/case_form.py
eatyourpeas: The form sets the none values in the model to empty strings. This might be because blank is True, so possibly my fault. If the child is from Jersey the URN is cleaned and the NHS number is set to none. I am happy to have some better approach. The clean_* methods are called before the final clean method, so the intention was to do some basic validation here first.

epilepsy12/management/commands/seed.py
eatyourpeas: Shazam!

epilepsy12/forms_folder/case_form.py
eatyourpeas: absolutely. i have put a `unique_together` field in the model.

epilepsy12/views/case_views.py
eatyourpeas: The ZZ codes are [NHS standards](https://digital.nhs.uk/data-and-information/data-collections-and-data-sets/data-sets/mental-health-services-data-set/submission-guidance/make-a-basic-submission#:~:text=If%20the%20postcode%20of%20the,Postcode%20is%20unknown%20%2D%20ZZ99%203WZ).  Jersey are not part of the NHS so I guess will not have any codes. NHS Wales have their own versions of the same but different codes. Maybe there is a code for people who live on yachts?

epilepsy12/models_folder/case.py
eatyourpeas: I agree

eatyourpeas:
A number of additional changes included. Main change is to add a `unique_together` constraint to the `Case` model for URN and NHS number. I have not added the 6 digit validator as we agreed today. Other bits all taken on board and amended.

mbarton (approved):
I've run this up locally, seeded some data and added patients in both Jersey and Kings. All seems to be fine, nice work!

eatyourpeas:
merging this will be quite complicated as there have been a bunch of migrations in between when this was created and now. Perhaps we could pair on this?

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 30 minutes and 8 seconds ago) Please check your changes!

PR #1127, opened 2024-12-07
optional-causes

### LINKED ISSUE #1125: Change the causes question on platform to "non-mandatory"
Hi Michael & Simon, 

Whilst we figure out a better process for adding causes onto the system can we please make the section non mandatory. 

So we would like to keep this question mandatory:
image

but then make the follow up question non-mandatory 

image

We have had a lot of queries come in so would ideally need this as soon as possible. 

Let me know if there is anything you need from me.

Thanks
Saleema

eatyourpeas:
### Overview

This makes the epilepsy cause drop down an optional field.

It adds epilepsy_cause to the list of fields that are nonessential and removes it from being included in the form calculation.
It also removes the blue target implying it is an essential unfilled field by introducing a new `allow_none` field in the `model_select` partial html element. This adds a `No choice made` choice in the select that returns an `empty` value in the htmx response that is used to save None to that field.

closes #1125

### DIFF
--- a/epilepsy12/common_view_functions/recalculate_form_generate_response.py
+++ b/epilepsy12/common_view_functions/recalculate_form_generate_response.py
@@ -197,9 +197,9 @@ def total_fields_expected(model_instance):
         # if a cause for the epilepsy is know other essential fields must be included
         if model_instance.epilepsy_cause_known:
             # essential fields include
-            # epilepsy_cause
+            # epilepsy_cause (# nolonger scored - issue #1125)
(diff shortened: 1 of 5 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @eatyourpeas 28 minutes and 43 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 29 minutes and 47 seconds ago) Please check your changes!

Issue #1150, opened 2025-01-06
patient coordinates none causes scatterplot to fail


eatyourpeas:
500 error due to patients with no location data unable to plot

PR #1152, opened 2025-01-07
Email admins on error

mbarton:
Re-add the default Django behaviour of emailing errors to ADMINS if DEBUG=False.

I'm going to use this as a lightweight error reporting system to try and proactively investigate errors like #1150 before users report them. It may also destroy my inbox with noise, in which case I'll turn it off again! But ideally we'd triage errors and fix or suppress.


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 20 minutes and 12 seconds ago) Please check your changes!

PR #1159, opened 2025-01-11
kpi-1

### LINKED ISSUE #1154: Fix user tracking on Case
When deleting a case we don't record who made the deletion. We propose to set updated_by, save, then delete.

### LINKED ISSUE #1156: Check calculation for KPI 1, and potentially others
related to #1139 

@eatyourpeas We think that a similar issue is happening for KPI 1, where maybe patients are being not scored for the measure if they have a none next to 'have they been seen' question.

These are the numbers on the live platform for KPI 1 and 2:
image
- would expect both to be out of 2505

Additionally, @anjaleesyangbo has found these differences remain once we've filtered out the incomplete records:
KPI denominator = 1474 (18 NA)
KPI 2 & KPI 9 denominator both = 1492

Please could you look into this, and then also check the MRI calculation as we also made changes to this mid way through the year. We can also discuss more in the meeting tomorrow.

eatyourpeas:
### Overview

Recalculate KPI 1
The addition of `paediatric_neurologist_input_achieved` or `paediatric_neurologist_input_achieved` fields introduced options where edge cases were not being correctly scored. 
This introduces a reworking of the measure and 39 additional tests to cover the edge cases

closes #1154

### DIFF
--- a/epilepsy12/common_view_functions/calculate_kpi_functions/score_kpi_1.py
+++ b/epilepsy12/common_view_functions/calculate_kpi_functions/score_kpi_1.py
@@ -29,40 +29,45 @@ def score_kpi_1(registration_instance) -> int:
         return KPI_SCORE["FAIL"]
 
     # check all fields complete for either consultant or neurologist
-    all_consultant_paediatrician_fields_complete = (
-        (assessment.consultant_paediatrician_referral_made is not None)
-        and (assessment.consultant_paediatrician_referral_date is not None)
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 29 minutes and 59 seconds ago) Please check your changes!

PR #1160, opened 2025-01-11
fix-mri-unscored

### LINKED ISSUE #1158: KPI 5 - updating measure scores
As discussed in the meeting today, please can the following be in updated in KPI 5:
image

Can line 57 be updated from not scored to failed?

Additionally whilst taking a look at the test measure folders I identified this at the top of the KPI 5 test script:
image

On line 17 should it ineligible cases are scored 2 and not 0? I don't think this is actually an changes with the script but thought it might be worth editing to ensure its correct.

eatyourpeas:
### Overview

In the KPI 5 calculation, if `mri_indicated` were scored true, and a date is missing, the measure should faily but currently returns `NOT_SCORED`

This has been fixed. The comment also has been corrected to return the correct value.

A new test has been added for this.

closes #1158


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 29 minutes and 59 seconds ago) Please check your changes!

PR #1175, opened 2025-01-17
Fix adding patients if postcode validation service is down

mbarton:
Currently findthatpostcode is returning a 429. In this case we should allow the postcode since we don't know whether it's valid or not.

Also fix the code that generates random postcodes for testing to handle the same error.

The code still emits an error calling coordinates_for_postcode:

```
django-1   | ERROR [epilepsy12.models_folder.case:218] Cannot get longitude and latitude for SE129QF: cannot unpack non-iterable NoneType object
Traceback (most recent call last):
django-1   |   File "/app/epilepsy12/models_folder/case.py", line 202, in save
django-1   |     lon, lat = coordinates_for_postcode(postcode=self.postcode)
django-1   |     ^^^^^^^^
django-1   | TypeError: cannot unpack non-iterable NoneType object
```

but that's wrapped in a try catch so doesn't block saving the patient

### DIFF
--- a/epilepsy12/general_functions/postcode.py
+++ b/epilepsy12/general_functions/postcode.py
@@ -26,12 +26,21 @@ def is_valid_postcode(postcode: str) -> bool:
 
     if response.status_code == 200:
         return True
+    
+    if response.status_code in [400, 404]:
+        # Log this at error so we still email the admins about it.
+        # This is to try and spot any examples where a correct postcode is marked as invalid.
+        logger.error(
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 29 minutes and 39 seconds ago) Please check your changes!

PR #1180, opened 2025-01-19
transfer-kpi-organisation

eatyourpeas:
### Overview

When users transfer patients between lead centres, a workflow uses the Site model to store where people are in the transfer process, as well what relationship the organisation has with the patient (whether it is the lead centre, responsible for general paediatric, neurology or surgical care). There is also a relationship with Organisation through the KPI table to allow the KPIs to be aggregated. 

A bug that recently came to light is that when the patient is transferred from one lead centre to another, the relationship between the patients KPI instance and the organisation is not updated.

This PR introduces this in the `case_views` file in the `transfer_response` function which is called when a user either accepts or rejects the transfer request from another centre.

There are a lot of permutations to this workflow which in retrospect might possibly have been differently. Tests are added in this PR for options where the request is accepted or where the request is rejected. There are still edge cases not captured, depending on whether centres after transfer remain actively involved with the patient in another capacity (general paediatrics / neurology / surgery).


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 13 minutes and 1 second ago) Please check your changes!

PR #1178, opened 2025-01-17
Fix submission on last day of grace period and close cohort 6

mbarton:
Fixes #1172

Undo the hot fixes we did to quickly unblock submission on the last day of the grace period (#1163, #1164, #1165) and the underlying bugs:

- `days_remaining_before_submission` should be inclusive of the last day of the grace period
- `submitting_cohort` was a number so we couldn't get `cohort_start_date` from it, we needed `submitting_cohort_start_date`

This will have the effect of closing out cohort 6 so will only merge after our extended deadline (Sunday 19th Jan).

Also removes the "0 days remaining" for cohort 6 on the homepage.

Finally as a side effect also fixes #1176

PR #1184, opened 2025-01-21
Fix edit button on unregistered cases for non audit team members

### LINKED ISSUE #1183: User unable to edit entry after adding patient
Hi, 

A clinician added a child to the platform (Case Number: 9898), but they were unable to edit or add any data. The FPA date had not been entered and the record was greyed out.

Image

I've managed to go in and add a FPA date, and she now is able to edit the entry. However when she goes to add another patient (9951) she experiences the same issue as above.

Can you look into this please? Do let me know if there is anything else you need from me. 

Thanks!

Saleema

mbarton:
Fixes #1183 

I introduced `Case.editable` in #1178 but it didn't work properly. `self.registration` was throwing `RelatedObjectDoesNotExist` for unregistered cases. The template rendering ate this somehow so the page still renders. Either way I didn't notice because I'm an `rcpch_audit_team_member` so can always edit.

This PR changes the logic to use `hasattr` and states out each expected branch separately.

### DIFF
--- a/epilepsy12/models_folder/case.py
+++ b/epilepsy12/models_folder/case.py
@@ -165,7 +165,13 @@ class Case(TimeStampAbstractBaseClass, UserStampAbstractBaseClass, HelpTextMixin
         return stringify_time_elapsed(self.date_of_birth, today_date)
     
     def editable(self):
-        return not self.locked and (self.registration and self.registration.days_remaining_before_submission > 0)
+        if self.locked:
+            return False
+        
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 30 minutes and 5 seconds ago) Please check your changes!

PR #1189, opened 2025-01-23
kpi-organisation-update-at-transfer

### LINKED ISSUE #1187: transfers not updating organisation in KPI if user neither accepts nor rejects request
When the transfer request is in flight (`update_lead_site`) the Site model is updated but not the KPI model with the new organisation. 

On accept/reject this does now happen but there is the possibility that a user may neither accept nor reject the patient, so there will be a mismatch

To fix this as well as update the Organisation in Site at the time the request is made, the same should happen for KPI, and reverse this only if the request is actively rejected.

eatyourpeas:
Fixes #1187 

### Overview

The transfer process not updating the organisation in the related KPI instance was fixed in a previous PR. This ensures that the KPI instance is updated at the point of transfer.

It also adds a test

### DIFF
--- a/epilepsy12/tests/common_view_functions_tests/registration_view_tests/test_registration_views.py
+++ b/epilepsy12/tests/common_view_functions_tests/registration_view_tests/test_registration_views.py
@@ -25,6 +25,8 @@ def test_tranfer_centre(
     client,
     e12_case_factory,
     e12_site_factory,
+    seed_groups_fixture,
+    seed_users_fixture,
 ):
     """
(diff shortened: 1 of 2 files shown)


mbarton (approved):
Nice one!

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 29 minutes and 7 seconds ago) Please check your changes!

PR #1191, opened 2025-01-28
Fix adding patients outside of Jersey

### LINKED ISSUE #1190: Users unable to enter new patients on the Platform
Hiya, 

I’ve received two separate queries from users who are unable to add new patients, encountering a message that reads: "Unique reference number (URN) already exists."

For the first query, I’ve checked, and there is no record of the patient in the system, nor has the patient been uploaded by any other user/trust. Therefore, they should be able to create a new entry...but they can't. 

For the second query, the user has confirmed that the patients are newly diagnosed and would not have been added to the platform previously by them or any other trust.

Could you please look into this?

Let me know if there is anything you need from me.

Saleema

mbarton:
Fixes #1190 

Patients outside of Jersey have `nhs_number`, not `unique_reference_number`. However they were ending up with their URN being the empty string (Django convention). This then broke the `unique_together` constraint.

This PR adds a cleaner to convert the empty strings to null, following the suggestion here: https://stackoverflow.com/questions/5772176/django-unique-together-and-blank-true.

We don't have the problem in reverse as we already null out the NHS number for Jersey patients in the respective cleaner.

### DIFF
--- a/epilepsy12/forms_folder/case_form.py
+++ b/epilepsy12/forms_folder/case_form.py
@@ -204,3 +204,18 @@ class CaseForm(forms.ModelForm):
             return formatted_nhs_number
         else:
             raise ValidationError(f"{formatted_nhs_number} is not a valid NHS number")
+    
+    def clean_unique_reference_number(self):
+        unique_reference_number = self.cleaned_data["unique_reference_number"]
+
+        organisation = Organisation.objects.get(id=self.organisation_id)
+        if organisation.ods_code == "RGT1W":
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @mbarton 28 minutes and 55 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 29 minutes ago) Please check your changes!

PR #1219, opened 2025-04-13
Issue879-multiple-employers

### LINKED ISSUE #879: Allowing users to join multiple trusts
We have a number of users (typically ICB/regional leads or staff working across Trusts) that would benefit from access to multiple sites.
They may also need different permissions at different sites (e.g. Lead at one organisation, but clinician at another). 

Currently, we're creating extra accounts for them using plus addressing.

eatyourpeas:
### Overview

This PR is a long-awaited feature that allows Epilepsy12 users to be affiliated with multiple organisations

Because the `organisation_employer` is stored as a foreign key in the `Epilepsy12User` model, the users have to be migrated over to the new table stepwise.

This PR therefore introduces 3 new migrations:
1. migration 0051: introduces a new OrganisationEmployer with foreign keys for both the Epilepsy12User and Organisation models
2. migration 0052: a custom migration that copies across all the foreign references to the new model if the existing `organisation_employer` field is not empty
3. removes the `organisation_employer` field in `Epilepsy12User` 

Step 3 replaces the `organisation_employer` field in the `Epilepsy12User` model  with an `@property` method of the same name to allow backwards compatibility and minimal change to the code base. It also adds an extra `set_organisation_employer` method for easy setting of new employers in future.

The other changes are the minimum refactor required to replace all preexisting function and include:
1. refactor of `create_superuser` and `create_user` to allow new users to have organisations in the new model
2. updates to `admin` (still needs more work)
3. refactors to the `epilepsy12_user_list` to allow filters and sorting to work and and primary organisations to appear in the table
4. refactors to some of the transfer site workflow where the Epilepsy12User model is filtered
5. refactors to the update lead site workflow as again the Epilepsy12User model is queried.
6. refactor to the user creation workflow where the new user is associated with an organisation

**The is the first step to introduce a new M2M table without disrupting existing function - the next step once merged would be to add workflows for adding and removing employers that are not primary**

## TODO

The main vulnerability of this are the migration steps, especially 0052: it seems to work for me and the logic behind it makes sense. There is a logging step to ensure that we retain the email and affiliation of any users that fail to migrate their organisation_employer.

Ideally we would deploy this first to staging and make sure that no user organisation affiliations were lost.


mbarton:
Tested locally:

- Starting a fresh instance, creating a superuser, adding a new user, logging in as the new user
- Take an existing instance with a superuser and two users in different orgs, then checkout this branch

The latter threw an error whilst trying to migrate users over to the new model:

```
django-1   | ERROR [epilepsy12.migrations.0052_migrate_users:33] Error creating OrganisationEmployer for user michael.barton@rcpch.ac.uk: 'OrganisationEmployer' instance expected, got <Organisation: Organisation object (139)> from Organisation object (139)
django-1   | ERROR [epilepsy12.migrations.0052_migrate_users:33] Error creating OrganisationEmployer for user kings@college.com: 'OrganisationEmployer' instance expected, got <Organisation: Organisation object (139)> from Organisation object (139)
django-1   | ERROR [epilepsy12.migrations.0052_migrate_users:33] Error creating OrganisationEmployer for user kingston@hospital.com: 'OrganisationEmployer' instance expected, got <Organisation: Organisation object (141)> from Organisation object (141)
```

eatyourpeas:
@mbarton very grateful for you doing this thank you. I think now fix applied and should now work.

mbarton:
Thanks - migrating existing users seems to work now.

I have a question about what happens when we merge this PR though. It removes the old `organisation_employer` field as the final migration and this will run at startup before the old instances are unregistered from the load balancer. I think that would cause errors in the period between that migration starting and the old instances being retired.

If we're confident that the new instances will come up without issue then maybe we can tolerate that?

I wonder if it's safer to move the migration removing the old field to a separate PR and merge that once we're up and running on new instances reading from the new model.

eatyourpeas:
The reason to remove the old field is that it will otherwise clash with the existing foreign key:
```
@property
    def organisation_employer(self):
        """Return the primary organisation for backward compatibility"""
        try:
            return (
                self.employer_organisations.filter(
                    epilepsy12_user=self, is_primary=True, is_active=True
                )
                .first()
                .employer_organisation
            )
        except AttributeError:
            return None

    def set_organisation_employer(
        self, organisation_employer, is_primary=True, created_by=None
    ):
        """Set a user's organisation, maintaining backward compatibility"""
        if organisation_employer:
            OrganisationEmployer = apps.get_model("epilepsy12", "OrganisationEmployer")
            org_emp, created = OrganisationEmployer.objects.update_or_create(
                epilepsy12_user=self,
                employer_organisation=organisation_employer,
                defaults={
                    "is_primary": is_primary,
                    "is_active": True,
                    "created_by": created_by or self,
                },
            )
            # If this is the new primary, make sure no other employments are primary
            if is_primary:
                OrganisationEmployer.objects.filter(
                    epilepsy12_user=self, is_primary=True
                ).exclude(pk=org_emp.pk).update(is_primary=False)
```
The `@property` field has the same name as the old field and this is to maintain backwards compatibility to prevent refactoring the code base. If both coexist they would be a clash. Should we try it on staging first and see if this actually happens or are there not enough users in there?

mbarton (approved):
Sorry I didn’t get to this before I left.

I don’t think the number of users matters and things were preserved locally for me when I tested.

The overlap between the new containers and old containers will be small and it’s the holidays.

I’m fine for you to merge this.

mbarton:
Overdue on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 30 minutes and 4 seconds ago) What's gone wrong?

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @eatyourpeas 30 minutes and 9 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 32 minutes and 11 seconds ago) Please check your changes!

Issue #1133, opened 2024-12-18
Upgrade to Postgres 16


mbarton:
Closes #1119 

No particular rush on this as Postgres 15 is supported until November 11, 2027. ~Upgrading would speed up our tests in CI though from 30 mins to 10 mins (see #1119)~ version didn't seem to matter here.

I've tested locally and it all seems to work. However I don't think we should do this before the end of January 2025 as that's the data submission deadline for cohort 6 and we don't want any risk before then.

Postgres 17 was released in September 2024 and is still in preview on Azure (https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-supported-versions). If it is out of preview by the time we get to this issue I would suggest we hop to 17, although that may mean longer downtime as you can only upgrade one major version at a time.

PR #1222, opened 2025-04-16
Upgrade to Postgres 17 for local development

mbarton:
Part of #1133. ~Should significantly speed up our CI run (from 30 mins to 10 mins).~ Postgres 16 didn't seem to, at least the first run here is as slow as before. Trying Postgres 17 now.


mbarton:
Looks like Postgres version was a red herring for the slow tests as I’m still seeing slow runs on 16 and 27

PR #1233, opened 2025-05-07
toggle selections from form

### LINKED ISSUE #1147: content security policy error in the console
This is an advisory error:
```
Content Security Policy 'upgrade-insecure-requests' was delivered via a <meta> element outside the document's <head>, which is disallowed. The policy has been ignored.
```
It occurs because the CSP should be in the head tag as a meta tag and not outside. The CSP in Epilepsy12 is in the head tag as normal, but several htmx requests return partials that extend the `base.html`. To fix this all the partials need to be amended to include only those parts of `base.html` that are needed, minus the head tag.

eatyourpeas:
### Overview

Fix for toggle to some of the facets in the form not being removed from the url on form submission.

### DIFF
--- a/epilepsy12/templatetags/epilepsy12_template_tags.py
+++ b/epilepsy12/templatetags/epilepsy12_template_tags.py
@@ -628,10 +628,6 @@ def strip(value):
     return value
 
 
-# ... existing code ...
-from django.http import QueryDict
-
-
 @register.simple_tag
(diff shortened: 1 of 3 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 11 minutes and 53 seconds ago) Please check your changes!

PR #1254, opened 2025-07-20
Issue1253-signals

eatyourpeas:
### Overview

This ports over the logging feature from NPDA
- updates the `signals.py` to notify significant changes in the `Epilepsy12User` and `OrganisationEmployer` models. As in NPDA, there are fields that trigger emails as well as logging.
- refactors `middleware.py` to support this
- adds some new options to the `VisitActivity` model. This generates a new migration
- refactors the user log and user table with the new visit activity fields
- enables reactivation of a user in the form and table for those users with delete permissions
- updates the client ip method to pick up `HTTP_X_FORWARDED_FOR`

#### A few things
1. Emails should be sent to all people in the `ADMINS` list. These should be pulled in from the env config in Azure but I don't seem to have the password to this so am not sure if it is configured. I have set a fall back of the Epilepsy12 team if `ADMINS` is empty and improved the `send_email_to_recipients` function to accept tuples as well as simple email addresses
2. The signals send also on changes to the `OrganisationEmployer` though this is not actually set up in the UI yet. Currently the dropdown filters only to the organisation of the user, and only audit team and superusers can view all organisations. This is a separate piece of work to enable, but the table structure allows users to have multiple employers (though only one primary).


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @eatyourpeas 10 minutes and 45 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @eatyourpeas 11 minutes and 49 seconds ago) Please check your changes!

PR #1259, opened 2025-07-24
Fix user signal emails

mbarton:
Follow on from #1254.

The Django ADMINS setting is a list of [name, email]:

```python
[("John", "john@example.com"), ("Mary", "mary@example.com")]
```

https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-ADMINS

We were using the entries in the list directly as the email for user notifications, causing a crash:

> Failed to send Employer assignment notification for user XXX.XXX@nhs.net: Invalid address "['michael', 'XXX@rcpch.ac.uk']"

I've added a new setting, `CHANGE_NOTIFICATION_EMAILS`. In dev and staging this will be empty but in live will contain the emails to notify.

This also stops us implicitly relying on both the `ADMINS` and the `SITE_CONTACT_EMAIL` setting for this behaviour.


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @mbarton 9 minutes and 45 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 10 minutes and 48 seconds ago) Please check your changes!

Issue #1237, opened 2025-05-29
patient scatterplot not plotting


eatyourpeas:
The map and the site render but not the patients

eatyourpeas:
unfortunately #1243 did not fix this

eatyourpeas:
will fix with #1263

PR #1263, opened 2025-07-27
complete-implementation-of-multiple-employers

### LINKED ISSUE #1261: UI for managing multiple employers
Follow on to #1219 to add the user interface for managing multiple employers and signposting to users they are members of more than one

eatyourpeas:
### Overview

Multiple organisation employers have been implemented for some time in E12, but there has been nothing in the user interface to allow this.

This PR implements:
1. A new route and template workflow for adding/removing employers and setting `is_primary` status
2. removes the organisation employer dropdowns from the create and update E12 user forms
3. updates the `user_may_view_this_organisation` decorator with a new `organisation_white_list_for_user` function which compares the organisation being accessed in the URL with all the organisations in all the trusts and LHBs with which the requesting user is affiliated.
4. Updates the dropdown in the `selected_organisation_summary` to include all organisations in the requesting user's employer list.
5. Refactors previous tests checking validation of organisation employer in the form to better end to end tests checking user creation through the view
6. Adds new tests to check users can access whitelisted organisations only
7. Updates the documentation

### Related Issues
Closes #1261

#### New User Form
image

#### Confirmation Modal
image

#### Edit Employer form
image

This shows the multiple organisations this user now has access to
image


epilepsy12/common_view_functions/sanction_user_access.py
```diff
@@ -29,3 +22,35 @@ def logged_in_user_may_access_this_organisation(user, organisation_requested):
     else:
         # user is not active or email confirmed
         return False
+
+
+def organisation_white_list_for_user(epilepsy12_user):
```
mbarton, line 27: Do we need this helper method rather than using `epilepsy12_user.employers.all()` directly? At least for the check in `user_may_view_this_organisation` if I do

```python
organisation_requested in user.employers.all()
```

it still seems to work

eatyourpeas: I think the `epilepsy12_user.employers.all()` with return all organisations that the user is affiliated with. But I had understood that the lead clinician needed access to all the organisations in the governing Trust or Local Health Board. The helper function runs a query to return all the sibling organisations of all the parents the user is affiliated, not just the organisation itself. The function is run in two places (in the view to populate the dropdown and here in the decorator), which is why I abstracted it out into its own function.

mbarton: Ah of course, thanks! I was still thinking of the NPDA permission model

nikyraja:
Thank you so much Simon and Michael! Those screenshots looks good, and I like how you've chosen to display this on the UI - very slick. 

The change will be really helpful for users who work across multiple Trusts, and for regional/ICB leads who want to view data submissions across their area.

Happy for you to make this live. We will include the change in our next comms.

mbarton (changes requested):
Thanks for putting this together @eatyourpeas. I ran it up to test and logged in as the automatically generated lead clinician.

I was able to edit my own employers and add any other organisation. Before we merge this we will need to lock that down, both in the UI and in the backend, so that only RCPCH audit team members can add to any organisation.

eatyourpeas:
Thanks @mbarton for reviewing. Sorry I thought that was the behaviour that was requested? It was only the deleting behaviour that I thought was locked down. I don't think I had appreciated that E12 were the only people that could allocate sites. I can have a look at that this weekend ?

nikyraja:
Thanks both! 

Apologies, we probably wouldn't have specified that giving users access to multiple orgs should be restricted to only the RCPCH team. However, I do think this would be sensible to add in, to prevent unauthorised access to multiple sites. Thanks for raising Michael, and thanks in advance for resolving both.

mbarton:
Thanks @nikyraja. FWIW in NPDA you can add users to any unit you are a member of (and RCPCH audit team can add to any unit): https://github.com/rcpch/national-paediatric-diabetes-audit/blob/db1ae089b8a19c299e735f155eca68803273eda0/project/npda/views/npda_users.py#L509.

I'm happy with following that in E12 or simply restricting editing the organisations for a user to the audit team.

nikyraja:
Ideally, users would still be able to add users to their own singular organisation, but access to multiple orgs would only be possible through the E12 team - would that be possible?

eatyourpeas (commented):
Thank you both - I have changed so that only 12 team can allocate employers

I have btw @mbarton added a new permission `can_allocate_user_to_organisation` which generates a new migration, and updated the decorators. There is also a new dev user (`e12_audit_team`) to make testing a little easier. There are additionally some tests in `tests/view_tests/permissions_tests/test_permissions_update.py`

mbarton (changes requested):
Adding a new employer for a user doesn't appear to work using the example users. As the superuser I tried to add Addenbrookes as an employer to myself and the demo lead clinician user but got the following error:

```
django-1   | ERROR [django.request] Internal Server Error: /organisation/139/epilepsy12_users/2/add_employer                                                django-1   | Traceback (most recent call last):                                                                                                             django-1   |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner                                          django-1   |     response = get_response(request)                                                                                                           django-1   |                ^^^^^^^^^^^^^^^^^^^^^                                                                                                           django-1   |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
django-1   |     response = wrapped_callback(request, *callback_args, **callback_kwargs)                                                                    django-1   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/app/epilepsy12/decorator.py", line 380, in wrapper
django-1   |     return view(request, *args, **kwargs)
django-1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/app/epilepsy12/decorator.py", line 174, in wrapper
django-1   |     return view(request, *args, **kwargs)
django-1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/app/epilepsy12/decorator.py", line 350, in wrapper
django-1   |     return view(request, *args, **kwargs)
django-1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/usr/local/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 59, in _view_wrapper
django-1   |     return view_func(request, *args, **kwargs)
django-1   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/app/epilepsy12/views/user_management_views.py", line 668, in add_employer_organisation
django-1   |     if request.user.permissions.filter(
django-1   |        ^^^^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/usr/local/lib/python3.11/site-packages/django/utils/functional.py", line 253, in inner
django-1   |     return func(_wrapped, *args)
django-1   |            ^^^^^^^^^^^^^^^^^^^^^
django-1   |   File "/usr/local/lib/python3.11/site-packages/django/utils/functional.py", line 253, in inner
django-1   |     return func(_wrapped, *args)                                                                                                               django-1   |            ^^^^^^^^^^^^^^^^^^^^^
django-1   | AttributeError: 'Epilepsy12User' object has no attribute 'permissions'
django-1   | 172.20.0.1 - incubator@rcpch.ac.uk [26/08/25:10:13:24 +0100] "POST /organisation/139/epilepsy12_users/2/add_employer" 500 107113 "https://e12.localhost/organisation/139/epilepsy12_users/2/add_employer" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36 Edg/139.0.0.0"
```

eatyourpeas:
OK so huge apologies for this. I am not really sure how I thought this was ready. Possibly those emerging grey hairs at my temples have started to grow inwards and are tickling my brain. The error was coming from an invalid attribute on the `request.user` where I had confused the permissions model with the `has_perm` attribute. It was not identified in the tests as the post request was missing the htmx header in the test, so the broken code would not have been called as it would not have been an htmx POST request.

## What I have fixed:
1. I have replaced the broken `request.user.permissions` with `request.user.has_perm(...)`
2. I have added a new partial template `add_employer_partial.html` to be included in `add_employer.html` and moved the target here. Now the conditional in the view checking if the request is an htmx one or not will return the partial template (without `base.html` etc as that is only present in `add_employer.html`). If the user has permission for edit access, this flag is set here also, and this will return a partial that will show/hide/disable buttons for editing the employer.
3. For clarity, users that appear in an organisation list where that organisation is NOT their primary, but they are also employed there, an info icon appears with context to explain this. (see screen shots)
4. I have corrected the test to add the htmx header so that the test best reflects real life. The tests check that e12 members can update employers to all roles. They also test that all roles except e12 audit members fail to update employers of all user types and raise a 403 denied. 
5. I have added similar tests for removing employers and setting primary on employers also. I have used the same permission, since it is e12 audit team, but could add a new permission for removing employers if needed.

## What to Check before merging:
1. log in as an e12 audit team member
2. Navigate to add / edit users in King's
3. edit the admin user record - the add employer button should be visible
4. add Addenbrooke's as a new employer - this should appear as a primary employer and KCH is demoted
5. Make sure you can set primary on king's again, delete addenbrooke's.
6. If these work, add addenbrooke's back as primary, kch second affiliation
7. return to the user table - set the view to organisation (rather than national)
8. the user should still be in the table with Addenbrooke's in the primary employer column (even though this is KCH organisation view), but next to an info icon that states they are still employed by King's
9. Logout
10. Login as the admin user
11. Check that the organisation drop down list on the organisation dashboard includes both addenbrooke's and king's
12. navigate to the users list and edit a user.
13. There should be no Add Employer button

## What to do after merging
On rebuild of the container the new migration will run and create the new permission but this will not automatically be added to the E12 audit team group. This will have to be done manually in the admin. Note that all seeding functions have been updated to include this permission in the correct groups for fresh installs.

Small Addendum:
[0c8be69](https://github.com/rcpch/rcpch-audit-engine/pull/1263/commits/0c8be69923b031ef950abb0571050554b6e755b3) did not actually fail - that one is on github. i reran the action and the tests now all pass. The red cross in the UI is old.

## Anything Else
Included in this PR is a fix for, and should close #1237. This is added in the issue as well.

image

image

image

mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (created by @eatyourpeas and merged by @mbarton 10 minutes and 55 seconds ago) Please check your changes!

Issue #1266, opened 2025-08-13
Reports of Platform being slow


SAbdin93:
Hi,

We’ve received a few emails reporting that the platform has been running slowly. I’ve noticed the same issue, logging in takes quite a while, and searching for trusts/patients or entering data is also slow.

Would you be able to look into this for us?

Thanks so much,
Saleema

mbarton:
Correlating this with data from Azure - during working hours the maximum response time does seem to spike up to around a minute, occasionally spiking higher

<img width="1506" height="449" alt="Image" src="https://github.com/user-attachments/assets/2fa9d62d-8953-4648-92d4-ac3ec371c495" />

CPU and memory usage on the container seem to be fine. 

CPU usage on the database is relatively high, with a 100% spike that seems to correlate with high latency spikes on the container

<img width="1501" height="355" alt="Image" src="https://github.com/user-attachments/assets/6f9a0fc7-4101-4544-b9c2-ae4a1ecb3615" />

The platform metrics for CPU credits consumed and remaining *seem* to indicate we're not being throttled.

Database memory usage is high but constant:

<img width="1502" height="453" alt="Image" src="https://github.com/user-attachments/assets/228d616d-345a-4c66-9e84-5a3623fc0e5f" />

We have 25GB remaining of the 32GB allocated so I don't think disk pressure is an issue yet.

We are currently running a `Burstable, B1ms, 1 vCores, 2 GiB RAM, 32 GiB storage` instance. I suspect that is small for our needs. I will try initially bumping it to `Burstable, B2ms 2 vCores, 4 GiB RAM, 32 GiB storage` to see if the extra core and memory help.

mbarton:
Recording for posterity that upgrading the database type took 10 minutes! 5 minutes through I started to worry that it was waiting to drain existing connections from the web app so manually terminated it. No idea if that sped it up. 10 minutes to adjust some burstable VM config!

mbarton:
Back up and running on the new database instance. Loading the org dashboard was still quite slow (#1092) but other pages seem to load fine. I will give it a few hours of organic traffic before having a look at the metrics again to see if the upgrade made any difference.

My no-proof total hunch theories for slower performance:

- Every update to a patient triggers cascading KPI writes in the database at all levels (org, trust etc)
  - Perhaps that starves out other requests?
- The org dashboard is slow (#1092)

[...]

PR #1273, opened 2025-08-20
Speed up templates using the organisation selector

mbarton:
On my travels with Django Silk (#1267) trying to reproduce #1266 I noticed the organisation_list.html template referenced .trust/.local_health_board which meant an extra database query per organisation. The classic N + 1 ORM query problem.

Using `select_related` to fetch them in the same query reduces endpoints with this pattern from ~1s locally to ~100ms.


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 10 minutes and 47 seconds ago) Please check your changes!

PR #1274, opened 2025-08-24
Only run aggregations viewable from a given organisation page

mbarton:
Before this change we ran all the aggregations for all the organisations with data, even though the organisation summary page would only show you data for your current organisation and above.

This should significantly speed up rendering the organisation KPIs in production where a lot of organisations have data. Our monitoring shows that endpoint taking ten times longer than other endpoints.

This also fixes an adjacent bug where clicking Publish would publish KPI data for all organisations. This didn't really matter though because we removed the ability to view it and I checked the logs and nobody has clicked the publish button anyway.

### DIFF
--- a/epilepsy12/common_view_functions/aggregate_by.py
+++ b/epilepsy12/common_view_functions/aggregate_by.py
@@ -248,33 +248,15 @@ Filter cases, run aggregations and store results in KPIAggregation models - ther
 
 def update_all_kpi_agg_models(
     cohort: int,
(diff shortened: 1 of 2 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @mbarton 9 minutes and 47 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 9 minutes and 53 seconds ago) Please check your changes!

PR #1275, opened 2025-08-24
Add an RCPCH organisation for testing

mbarton:
Supports testing the live system without using a real organisation and thus granting access to patient data for that organisation


mbarton:
We didn’t need this for the pen test in the end, we used an empty existing organisation.

We can reopen and merge if we need it, however we’d need to filter out the patients from the KPI export and in analysis queries from direct database access

PR #1276, opened 2025-08-26
Case insensitive email for password reset

### LINKED ISSUE #1269: Password reset after expiry email is case sensitive
When you password expires we send you into the password reset flow where you have to enter your email address again. A user has reported that they are then told their account doesn't exist with that email.

This happens because the email field in the password reset flow is case sensitive. Their email is stored lowercase in E12 but they have it with uppercase first letters in their email signature, so I suspect they are entering that one in the reset field.

The login form itself is not case sensitive so unaffected.

mbarton:
Ronseal. Fixes #1269

### DIFF
--- a/epilepsy12/views/user_management_views.py
+++ b/epilepsy12/views/user_management_views.py
@@ -680,7 +680,7 @@ class ResetPasswordView(SuccessMessageMixin, PasswordResetView):
 
     # extend form_valid to set user.password_last_set
     def form_valid(self, form):
-        email = form.cleaned_data["email"]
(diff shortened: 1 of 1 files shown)


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/) (merged by @mbarton 9 minutes and 50 seconds ago) Please check your changes!

mbarton:
Seen on [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 10 minutes and 4 seconds ago) Please check your changes!

PR #1282, opened 2025-09-09
Delete historical record on deletion

mbarton:
Without this we retain demographic information for patients in the database after deletion even though it is no longer accessible via the platform


mbarton:
Seen on [STAGING](https://e12-staging.rcpch.tech/), [LIVE](https://e12.rcpch.ac.uk/) (merged by @mbarton 9 minutes and 41 seconds ago) Please check your changes!

PR #1318, opened 2025-11-14
Report-builder-rejig

### LINKED ISSUE #1317: Report Builder Items
Sprint 14/11/2025

Generally the team are happy that the performance of the report builder is faster. E12 team would like to make the following changes before removing the feature flag.

- [x] move the reset and add filter buttons to be near the active filters alert
- [x] if possible add labels to the kpi filters to remind users which kpi is which
- [x] add kpi reference text as popups to kpis
- [x] remove the 'Trust' and 'Local Health Board' labels from the dropdown to allow searching
- [x] downgrade access permissions to view for report builder and remove the guard around the anchor button

eatyourpeas:
### Overview

closes #1317

styling changes to the report builder prior to publishing

Styling changes include:
1. moving demographics filters in to national view filters segment (national view items will not render for regular users)
2. move the remove filters button into the filters alert
3. move the apply filters into the filters segment
4. fix the registered totals missing in the stats
5. add the kpi label to the item
6. add the kpi reference text to the tooltip for KPIs
7. remove the guard from the report builder anchor button to be visible to all users
8. downgrade the permission required to access the CaseFilter list view to `view_case`.
9. remove the `Trust` and `Local Health Board` text prepended in the local health board/trust dropdown in the national filter

That's it ready to publish

image

image

image

image

### DIFF
--- a/epilepsy12/constants/kpi.py
+++ b/epilepsy12/constants/kpi.py
@@ -64,12 +64,12 @@ KPI_MAP = {
     "10": "school_individual_healthcare_plan",
 }
 KPI_LABEL_MAP = {
-    "1": "Paediatrician with expertise in epilepsies",
-    "2": "Epilepsy specialist nurse",
+    "1": "Paediatrician with expertise in epilepsies within 2 weeks",
+    "2": "Access to Epilepsy specialist nurse",
     "3": "Tertiary input",
     "3b": "Epilepsy surgery referral",
     "4": "ECG",
-    "5": "MRI",
+    "5": "MRI within 6 weeks",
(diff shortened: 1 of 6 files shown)


nikyraja:
Thanks so much Simon! 

Only minor changes needed to the KPI titles please, if possible:

1 - Paediatrician with expertise within 2 weeks
2 - Access to Epilepsy Specialist Nurse
5 - MRI within 6 weeks

I can't see KPI 9, so just checking they are:
9a - Care planning agreement
9b - Care planning components

Otherwise, it looks great and thanks for making those changes so quickly! Happy for this to be made Live, and we will add to our comms.

PR #1328, opened 2025-12-11
valproate-pink-tick

### LINKED ISSUE #1300: User unable to submit patient with sodium valproate
Hiya,

For patient ID 6344, the user is trying to submit the entry but its saying 101% . Page 7 is not highlighted in pink so not sure if the issue is related to Sodium Valproate. 

Image

Could you please look into this when possible?

Thanks
Saleema

### LINKED ISSUE #1321: Can't add a medicine twice
Add a medicine with a discontinued date. Then try to add a new course of the same medicine. It doesn't let you select the same name again.

Reported by John Radcliffe Hospital (Oxford University Hospitals).

### LINKED ISSUE #1327: Valproate pink tick appears when record still incomplete
As in this example - the counter correctly reads that the record is incomplete (by 2 items) but the pink tick still shows for valproate. If you click into valproate the two unscored items are seen.

Image

Image

What should happen:
The valproate item should have a blue target until all the items are scored

eatyourpeas:
### Overview

Fixes two issues:

1. Antiepileptic medicines that had been discontinued (had a start and a stop date) were being excluded from the drop down if users subsequently wanted to add the medicine back. 
2. if valproate (or topiramate in cohort 7) were selected, the pink tick would appear after  the start date, safety declaration and medicine were selected. The additional fields (PPP and annual risk assessment form) were not included in the decision to show the pink tick. This meant users might not have completed the form but saw only pink ticks so had no visual cues as to where the missing fields were.

Code changes
1. `get_medicine_choices` helper now excludes only active medicines
2.  `medicine_id` resets the PPP and valproate risk acknowledgement to None on every call.
3. adds logging for future if form scoring finds score > expected
4. tidies some of the logic around PPP eligibility

image

image

closes #1327 
closes #1321
closes #1300

### DIFF
--- a/epilepsy12/common_view_functions/recalculate_form_generate_response.py
+++ b/epilepsy12/common_view_functions/recalculate_form_generate_response.py
@@ -1,5 +1,6 @@
 from dateutil import relativedelta
 from datetime import date
+import logging
 
 # 3rd Party Imports
 from django_htmx.http import trigger_client_event
@@ -27,6 +28,8 @@ from epilepsy12.constants import (
     Management_minimum_scorable_fields,
     AntiEpilepsyMedicine_minimum_scorable_fields,
 )
+
+logger = logging.getLogger(__name__)
 
 
 def recalculate_form_generate_response(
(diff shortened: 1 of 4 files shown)


epilepsy12/common_view_functions/recalculate_form_generate_response.py
```diff
@@ -86,6 +89,12 @@ def update_audit_progress(model_instance):
     else:
         registration = model_instance.registration
 
+    # Logging step
+    if all_completed_fields > all_expected_fields:
+        logger.error(
```
mbarton, line 94: Just commenting for posterity that this is conceptually a `logger.warning` but we are deliberately doing `logger.error` to trigger one of the Django error emails to admins

eatyourpeas: Good point. I actually started off with warning and changed it to error. Ironically that was the error. We should put money on whether you will actually get an email now that the PR has merged...

PR #1341, opened 2026-01-15
Updates KPI 8 (both <cohort 7 and after) to include only children where topiramate/valproate was prescribed within the first year of care. Excludes also children who are not girls (ie where sex is not known or not specified) where sex is important for the measure

### LINKED ISSUE #1339: Error with updated KPI 8
Hey Michael, 

Thanks for all your help today. Apologies but another issue has popped whilst doing some checks of the updated KPI 8 numbers.

I compared the num/denom for KPI 8 calculated by the platform was 57/66 but when I calculated it manually I got something closer to 43/48.

I managed to work backwards and ascertain that the differences in code were I was filtering == "female" as opposed to != "male" (platform) and that I had an additional parameter which filtered for AEDs that were only started in the first year of care (which I also don't think is specified in the platform code e.g. when I checked an individual case on the platform website, start date of topiramate was after first year of care had passed ). There was also an additional case which had NAs for the has_a_valproate_annual_risk_acknowledgement_form_been_completed & is_a_pregnancy_prevention_programme_in_place variables, which we believe is from before the KPI was updated. We believe the platform is correctly filtering this out.

I have spoken to Niky and she agrees that the code in the platform should be updated so only "female" are eligible as opposed to "not male" and as as the audit is only first year of care AEDs started after the first year of care should not be included. Happy to jump on a call tomorrow if required!

I have attached the file path to the KPI 8 calculations:
epilepsy12/common_view_functions/calculate_kpi_functions/score_kpi_8.py

Best wishes,

Anj

eatyourpeas:
## Overview

KPI 8 tracks the prescription of valproate in girls of any age, or topiramate in girls > 12y

Currently the measure is scored even if either of these medicines is prescribed beyond the first year of care. Patients are only supposed to be eligible for the measure if at least one of these medicines was started during their first year of care.

Also, through a clumsy conditional, patients that are not boys have been scored as girls (so including those scored as 'not known' or 'not specified').

This PR:
In kpi-8 calculation:
1. excludes all children that are not explicitly scored as female for KPI 8 (both pre and post cohort 7)
2. excludes all children that are on valproate/topiramate where it has been prescribed outside the first year of care
In the tests:
Adds a test for each of these possibilities

Documentation
The docstrings and the documentation have been updated to reflect this


mbarton:
I have recalculated the affected registrations by running this code manually in the container:

```python
regs = Registration.objects.filter(cohort=7)
for reg in regs:
    print(reg.id)
    sodium_valproate = KPI_SCORE["NOT_SCORED"]
    if hasattr(reg, "management"):
        age_at_first_paediatric_assessment = calculate_age_at_first_paediatric_assessment_in_years(reg)
        sodium_valproate = score_kpi_8_topiramate(reg, age_at_first_paediatric_assessment)
        reg.kpi.sodium_valproate = sodium_valproate
        reg.kpi.save()
```

It now shows 43/48 at national level as expected.

PR #1343, opened 2026-01-16
Organisational Audit - fix submit button not appearing

### LINKED ISSUE #1342: Org audit flip flop between submitted and not submitted
1. Create a 2024 org audit submission period and open it
2. Fill out the org audit
3. Close the 2024 period
4. Create a 2025 period and open it
5. Change just 4.1i (from yes to no or vice versa)

Expected: the progress indicator shows the option to submit
Actual: the progress indicator shows "complete" (what you see after you submit)

This is because we include a hidden `submitted` `input` element when all questions are complete. The questionnaire is "complete" because it has answers from last year - so the act of editing one of the questions submits the form including the hidden element which then marks the questionnaire as complete.

As it's complete, we don't include the hidden element and so if you change another answer it goes back to showing the button. This is expected behaviour. However changing the question answer again then puts it back to "complete"!

I think the hidden input is not the right solution here. Perhaps we could sniff `hx-trigger` on the server side to distinguish between a question being answered (which should always set `submitted=false`) and the submit button being pressed.

### LINKED ISSUE #1344: Org audit shows complete even though submit not pressed
In 2025 the following units were showing as "Complete" in the UI but their `submitted` flag was false:

- EPSOM AND ST HELIER UNIVERSITY HOSPITALS NHS TRUST
- NORTH WEST ANGLIA NHS FOUNDATION TRUST
- NORTHAMPTON GENERAL HOSPITAL NHS TRUST
- NORTHERN LINCOLNSHIRE AND GOOLE NHS FOUNDATION TRUST

It seems that the submitted flag was incorrectly copied across from the previous submission.

mbarton:
Fixes #1344
Fixes #1342

There was a silly bug where the submit button doesn't appear on the next years submission once complete if the previous year had a submission - it appears to be already submitted.

Debugging this was a bit of a nightmare as the `submitted` value used to control whether the submit button appeared came direct from the submission instance which was definitely `False` in the database.

However, calling `field.errors` when iterating through the fields on the `OrganisationalAuditSubmissionForm` would copy `submitted` from the initial data passed to it. This happened deep in the bowels of `group_form_fields`.

Also fixes flip flopping between submitted states by only setting `submitted=True` if the user for sure clicked the submit button.

### DIFF
--- a/epilepsy12/views/organisational_audit_views.py
+++ b/epilepsy12/views/organisational_audit_views.py
@@ -73,7 +73,7 @@ def group_questions(fields_by_question_number):
 
 
 def group_form_fields(form):
-    # Loops trough the form fields and groups them by section and question number
+    # Loops through the form fields and groups them by section and question number
     # The question number is taken from the help text of the field
     # This function is called with every field updated in the form and rerenders the form partial
 
@@ -182,8 +182,11 @@ def get_submission(submission_period, group, group_field):
 def get_submission_form(submission, last_submission):
     data = {}
     
+    fields_not_to_copy_from_last_year = OrganisationalAuditSubmissionForm.Meta.exclude
+    fields_not_to_copy_from_last_year += ["id", "submitted"]
+
     for field in OrganisationalAuditSubmission._meta.fields:
-        if field.name != "id" and not field.name in OrganisationalAuditSubmissionForm.Meta.exclude:
+        if not field.name in fields_not_to_copy_from_last_year:
             field_value = getattr(submission, field.name) if submission else None
             
             if not field_value and last_submission:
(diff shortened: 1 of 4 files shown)


epilepsy12/views/organisational_audit_views.py
```diff
@@ -237,6 +240,8 @@ def _organisational_audit(request, group_id, group_model, group_field):
 
         if number_completed != total_questions:
```
mbarton: I don't think this is the right approach. I'm not really sure why I had to add the `else` case here in the first place - I guess because it's excluded from the form now?

Probably makes sense to fix #1342 at the same time and drive this entirely from the action (i.e `hx-trigger`) in the UI rather than infer it from the completed question count

mbarton: Didn't need `hx-trigger` in the end, could just use `name`, `value` set on a `button` instead rather than `input[type="submit"]`

Issue #1347, opened 2026-01-30
Errors with Cohort 7 KPI export from platform


anjaleesyangbo:
Hi Michael,

We were doing some checks against our calculations and the KPI export from the platform and noticed a couple of major discrepancies with the KPIs for both England and Wales. I have attached a copy of the latest KPI export excel which includes the comparisons in a bit more detail.

1) There is a discrepancy in the total numbers of completed cases / uncompleted cases, I think due to a discrepancy in England. All registered cases for England don't match - 3544 (export) vs 3545 (our value). See excel tab 1 (Registered vs Total - Country). This then subsequently creates a discrepancy in the denominator in the KPI calculations at England level - however, the numerators from the KPI export match what we calculated see Excel tab 2 (Country_level).

2) The KPI numbers in the KPI export for Wales are very strange... the total number of completed cases for Wales is 164 but for the KPI export the numbers are very low e.g. 4/6 for KPI1 when our calculations show it should be more like 28/164. See excel tab 2 (Country_level). When checking the historical KPI exports (taken from mid Dec 25 to update the PowerBI) the numbers for Wales were much higher and more in line with what we would expect - suggesting that this bug has occurred since then.

3) When attempting to investigate the discrepancy in the England denominators, I went to look at the NHSE region totals to try and narrow down where the case might be located and once again identified some big discrepancies in the KPI numbers for EoE, SE and SW (the other regions appear to be correct). Similarly to point 2, these numbers appeared to be fine in the dec KPI export. See excel tab 3 (NHSregion_level).

Anj

[kpi_export_cohort_7-checks.xlsx](https://github.com/user-attachments/files/24964759/kpi_export_cohort_7-checks.xlsx)

mbarton:
Thanks for reporting @anjaleesyangbo - I am looking into this now

mbarton:
Confirming the country KPI aggregation object in the database for Wales is wrong:

```
>>> wales = Country.objects.get(name="Wales")
>>> wales_kpi = CountryKPIAggregation.objects.get(abstraction_relation=wales, cohort=7)
>>> wales_kpi.paediatrician_with_expertise_in_epilepsies_passed
4
>>> wales_kpi.paediatrician_with_expertise_in_epilepsies_total_eligible
6
>>> wales_kpi.paediatrician_with_expertise_in_epilepsies_ineligible
0
>>> wales_kpi.paediatrician_with_expertise_in_epilepsies_incomplete
0
```

[...]

PR #1354, opened 2026-02-07
Fix incorrect KPI calculations for cases with referrals outside of their org

mbarton:
Part of #1347 

#1274 had the unintended side effect of leaving incorrect KPI aggregation values in the database.

Imagine a case in Wales with a general paediatric referral to an organisation in England. When a user looked at their dashboard for any organisation in England, that case was incorrectly included for KPI calculation. The result was that the Welsh KPI numbers were pulled down.

You only noticed this through the KPI spreadsheet export as visiting any Welsh org would recalculate the values correctly. In the live system I don't believe there were any referrals the other way.

However the same bug exists at all abstraction levels, so a patient referred out of their trust could lead to incorrect numbers. Indeed we saw this at NHS regional level with lower than expected numbers in the South East.

There were already filters attempting to only include cases with active sites however they weren't working. I moved them to be after the filter by trust/icb/region etc in the list and it then generated a correct query. Go figure, looks like ordering matters in a Django filter! (Since [Python 3.6 at least](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep468)).

I've also added a unit test so whilst that's a fairly brittle fix, the test should catch regressions.


mbarton:
I'm expected this to fail as there is a previous test leaking case instances

PR #1355, opened 2026-02-07
Remove fixture seeding cases

mbarton:
It's hard to write e2e tests for KPI calculations (eg #1354) if there are existing cases in the database.

This PR replaces all usages of the old case seeding function with generating cases using the factory that only live for the lifetime of each test.


epilepsy12/tests/view_tests/permissions_tests/test_custom_permissions.py
```diff
@@ -1,102 +0,0 @@
-"""
```
mbarton, line 1: I doubt we are going to ever unskip this so I've removed it

PR #1368, opened 2026-04-18
removes the snomed causes from multiaxial diagnosis

### LINKED ISSUE #1224: Remove causes list for platform
Hiya, 

Our Methodology Dataset Group has agreed that the causes list should now be updated to reflect the list found here:

[nhsgms-panelapp.genomicsengland.co.uk/panels/402/v7.0](https://nhsgms-panelapp.genomicsengland.co.uk/panels/402/v7.0)

There are currently 610 entities so it would be good to run these as a look up from source.  We do not need break it down into the type of entity.

If possible, could we also include a reference in the code to indicate where the list is being sourced from? That would be really helpful for transparency.

Lastly, can the question be made mandatory again on the platform, please?

Hope this makes sense! 

Thanks! 
Saleema

eatyourpeas:
### Overview

Epilepsy causes are an ongoing discussion. The E12 audit team for the moment have decided to deprecate the SNOMED refset of causes that is seeded in the early migrations straight from our SNOMED server into the E12 database.

This PR does not change the seeding step (that now really only happens in development) nor does it remove any of the code and documentation around causes for the moment.

Instead it:
1. removes the view reference to epilepsy causes and associated htmx urls that call back on change of the dropdown, and passes the epilepsy causes from the table to the template.
2. removes those same views from `urls.py`
3. removes the epilepsy causes dropdown partial from the template in multiaxial diagnosis

This does not have any implications for form scoring since it was not a mandatory field in any case.

Tests updated to remove epilepsy causes.

closes #1224

### DIFF
--- a/documentation/docs/development/dropdowns.md
+++ b/documentation/docs/development/dropdowns.md
@@ -51,7 +51,10 @@ eg: `extra_causes_without_concept_ids = [...., {"preferredTerm": "Cause name", "
 
 Note that the function expects a list, even if only one item is supplied.
 
-<!-- There will need to be further documentation added here for new organisations and trust, as well as new comorbidities, and possibly medications and so on. For now, this is the workflow for EpilepsyCauses -->
+> [!NOTE]
+> Epilepsy Causes have as of #1224 been deprecated pending a better solution.
+
+
 
 ### Organisations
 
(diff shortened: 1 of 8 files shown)


mbarton (approved):

eatyourpeas:
@mbarton I now realise that if we are making epilepsy causes backwards compatible (ie viewable for historical cohorts) we might need to revert this. It feels a lot of fiddle for something we will never really use. What do you think? It seems to me there is an argument just to leave the current implementation of causes for all cohorts and remember that in cohort 6 some people filled in snomed causes from the database and if we need it we go back to the snapshots?
Alternatively, we would revert this (or cherry pick the commits into the audit period branch and add gating for cohort 6 and below)
What do you think?

PR #1367, opened 2026-04-16
2025 imd

eatyourpeas:
refactor(imd): move IMD calculation out of Case.save() into signal-driven utility

closes #1302

## Motivation

`Case.save()` called `imd_for_postcode()` directly, but the correct IMD
year depends on `Registration.cohort`, which may not exist when a Case is
first saved. This caused a `ValueError` ("Model instances passed to related
filters must be saved") and incorrect IMD lookups when the Registration
arrived after the Case.

### Core architecture
- **`epilepsy12/models_folder/case.py`** — stripped all IMD logic from
  `Case.save()`. The method now only normalises the postcode and calls
  `coordinates_for_postcode`. Sets `index_of_multiple_deprivation_quintile`
  to `None` when postcode switches to an unknown/placeholder value.
- **`epilepsy12/general_functions/index_multiple_deprivation.py`** 
  - added year parameter to `imd_for_postcode(postcode, year=imd_year)`: **NOTE** env must be updated with new census api URL
   — added`recalculate_imd_for_case(case)`: single source of truth for IMD
  calculation. No-ops when postcode is missing/unknown or Registration/cohort
  is not yet available. Derives `imd_year` from cohort (≥8 → 2025, else
  2019), calls `imd_for_postcode(postcode, year=imd_year)`, and persists via
  `queryset.update()` to avoid re-triggering `Case.save()`. Also fixed a
  stray `?` in the API URL.
- **`epilepsy12/signals.py`** — added four signal handlers:
  - `_capture_case_postcode` (pre_save Case)
  - `_recalculate_imd_on_case_save` (post_save Case) — fires when postcode changes
  - `_capture_registration_assessment_date` (pre_save Registration)
  - `_recalculate_imd_on_registration_save` (post_save Registration) — fires when assessment date changes

### Bulk recalculation tooling
- **`epilepsy12/management/commands/recalculate_imd.py`** — new management
  command (`manage.py recalculate_imd --all` / `--cohort N`) for
  backfilling IMD on existing records after a cohort boundary change or API
  update.
- **`s/recalculate-imd`** — convenience wrapper script.

### Tests
- **`epilepsy12/tests/model_tests/test_case.py`** — updated patch targets
  from `epilepsy12.models_folder.case.imd_for_postcode` to
  `epilepsy12.general_functions.index_multiple_deprivation.imd_for_postcode`;
  updated `assert_called_once_with` calls to include `year=2019`.
[...]

PR #1370, opened 2026-04-21
Revert "2025 imd"

mbarton:
Reverts rcpch/rcpch-audit-engine#1367

Reverting this out to investigate as it seems like the new signal is not firing to save IMD


mbarton:
I'm a 🍜 - the calculation is done at registration time! Cancelling this revert

PR #1382, opened 2026-05-13
fix/invalid-date-validation-genetic-tests

### LINKED ISSUE #1381: Genetic Testing Question - Issue
Hiya,

We’ve had an issue reported with case number 12140 relating to the genetic testing question.

The user is trying to enter the results obtained date, but the system is not allowing it even though the obtained date is after the requested date.

Requested date: 18/07/25
Obtained date: 26/08/25

Could you please have a look when you get a chance?

Image

Thanks!

Saleema

eatyourpeas:
### Overview

Current behaviour:
The R14/27/59 genetic tests in the Investigation view incorrectly tag the earliest date as `is_earliest_date`, instead applying it to whichever date is sent to the function by htmx, so that a meaningful comparison is never made. 

Ths PR used a TDD approach to fix the issue:
- ran 6 tests for expected behaviour (valid dates, invalid dates) one for each test to demonstrate failure
- fixed the `is_earliest_date` logic so that all test went green
- add guard path tests for invalid date types 

fixes #1381

### DIFF
--- a/epilepsy12/tests/view_tests/db_actions/test_genetic_testing_date_validation.py
+++ b/epilepsy12/tests/view_tests/db_actions/test_genetic_testing_date_validation.py
@@ -1,0 +1,175 @@
+from datetime import date, timedelta
+
+import pytest
+from django.urls import reverse
+
+from epilepsy12.models import Epilepsy12User, Organisation
+from epilepsy12.tests.UserDataClasses import test_user_rcpch_audit_team_data
+from epilepsy12.tests.factories import E12CaseFactory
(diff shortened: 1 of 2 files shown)


eatyourpeas:
This is fixed, tests passing but i have not been able to test it in live. I would be grateful @SAbdin93 if you could try it out and reopen the issue if it is still not working as expected?

**UPDATE**
image

mbarton:
Thanks for fixing @eatyourpeas! I really should have been testing this feature with realistic dates

PR #1385, opened 2026-05-29
fix(date_field): remove Set button, auto-submit on valid date

### LINKED ISSUE #1349: Confusion about when dates save
Via @coldunk there was a query at the BPNA if it was possible to save dates without clicking the "Set" button. It can be easy to forget to do until you're used to the platform (I also got confused by this).

@eatyourpeas explained that this was added because otherwise it would save before people had finished typing the date. There is already quite a lot of signposting to say the field is incomplete but we could perhaps include that on the button somehow?

My personal feeling is that the HTMX model breaks down a little for something like this - as soon as there's a valid value in the field we should save it but leave the cursor in the same place. That might mean an out of band call which doesn't result in any re-rendering on the page? Gut feeling though and I suspect we tried variations on this originally and they didn't work so feel free to ignore me

eatyourpeas:
## Summary

Removes the manual **Set** button from the date field component and replaces it with automatic HTMX submission using the native HTML5 Constraint Validation API.

The POST now fires on `change` only when `event.target.validity.valid` is `true` — whether the date was entered via datepicker or keyboard. This restores consistency with all other auto-submitting fields in the audit.

## Changes

- Remove `<button>Set</button>` and its Hyperscript from `date_field.html`
- Move all `hx-*` attributes and `hx_confirm` Swal confirmation onto the `<input>` directly
- Remove `hx_preserve` support (no longer needed)
- Remove unused Semantic UI mini modal
- Add optional `min_date`/`max_date` template parameters; default `min` of `1900-01-01` prevents premature submission on partial year entry (e.g. typing `12/05/0002` into the year field)
- Fix trigger filter expression: `event.target.validity.valid` (not `this.validity.valid` — HTMX evaluates filter expressions with `window` as `this`)

## Notes

- `registration_dates.html` is unaffected — the Swal confirmation modal fires via `htmx:confirm` before any request regardless of trigger type, so the confirmation flow is preserved without needing the button
- Callers can pass `min_date` to tighten validation further (e.g. clinically meaningful date floors)

Closes #1349


mbarton:
This works for me locally but I'm wondering should we keep the existing "set" button for the first paediatric assessment date?

If you type in a date into that field manually then the confirmation modal pops up as soon as the date is valid which is potentially a bit frustrating? I guess most people will probably use the date picker and click a particular date anyway which is fine.

I worry that since we don't allow people to change the fpa date this change might make it impossible to correct a mistake?

eatyourpeas:
yes i did wonder this but the modal does have a cancel button so it is possible to cancel out if the user has entered a date they did not mean to. And the admin team can end it I think. Also, I do think it is confusing (and has been confusing) that we promote 'save as you go' as part of the design but break it with dates.

Very happy to do this though. @nikyraja - what do you think

nikyraja:
Tahnk you both. Really pleased that we have introduced auto save into the date fields and think this will improve user experience. 

I am abit cautious about removing it from the first paediatric assemsent date though, because once the date is set it can't be edited by the users? In the past we had several users email to ask us to edit the fpa date, and because that produced alot of workload for the team, we introduced the pop-up prompt which works well. 

Another option is we make the users abe to edit the date, however I think a past reasoning was that this make things 'messy'..?

Apprecaite its abit contradictory, but I would therefore prefer to keep the 'set' button for the fpa date only alond with the pop up. Hopefully okay to have just the fpa date as the exception.

eatyourpeas:
**UPDATE**
Introduces `show_set_button` flag into `date_field` web component. Defaults to `False`

## With Set Button
image

## Without Set Button (default)
image

PR #1387, opened 2026-05-29
Patient level incomplete fields

### LINKED ISSUE #1386: add a list of missing measures
Once there are fewer thank 5 items to score, this should be signposted at the patient level
discussed at audit meeting 29/5/2026

eatyourpeas:
Signposts the number of incomplete measures and in which category back to users in the patient level list
image
closes #1386


mbarton:
@nikyraja can you confirm where you wanted this signposting to go? It's currently a tooltip on each row of the patient list (as per the screenshot above).

nikyraja:
I think the tooltip works well and looks good - thank you!

Could we also get this within the patient record, eg, below the % dial? But because we don't have question numbers, this might not work/look odd?
image

eatyourpeas:
Just an update here - I have started to implement a tooltip for form level completion status: 
image

For related models in the Multiaxial Diagnosis and Management (episode, syndrome, comorbidity etc) I have implemented a nested list eg:
> Tooltip
 *Episode 1:*
 When did the first episode of this type happen?
 *Episode Date*
 Episode 2:
 Description 

In the process I realised that the fields are not cohort-gates, so the newer fields in multiaxial diagnosis (eg relating to Autism Spectrum Disorder) and the deprecated epilepsy causes apply to all cohorts. I have put a list of affected fields in the docs and posted in the related issue #1391

Before completing this we should have an agreed list of measures which are visible only in cohort 6 vs 7 vs 8. I can apply that logic in this PR as part of this UI fix, and then any new AuditPeriod model additions could sit on top of this?

eatyourpeas:
so in anticipation i have put the epilepsy causes back in but gated it behind a cohort conditional - i have made a branch off this one here just in case this was not the final decision - so we can either merge this in now as a complete feature, or we could merge in the https://github.com/rcpch/rcpch-audit-engine/tree/replace-epilepsy-causes-conditionally branch into this one before merging it all into live

PR #1402, opened 2026-06-04
de-snomedify

### LINKED ISSUE #1394: Migrate off HERMES
We make SNOMED queries to our HERMES server during seeding to set up epilepsy causes, comorbidities and the medicines list.

We'd like to move off it for a few practical reasons:

- The server is quite old and we don't have the team resources to keep it up to date
- Relying on an external service during migrations means setting up dev instances and worse our deployment pipeline is dependent on that server being up

@pacharanero has made a really cool tool called [sct](https://github.com/pacharanero/sct) that would allow us to run the SNOMED queries without an additional server. However it does need to download the underlying data from the NHS TRUD and of course that would be even worse to rely on for our deployment pipeline!

So I think we need to invert how this works. First we need a centralised place where we download a copy of the SNOMED release from TRUD following [these instructions](https://github.com/pacharanero/sct/blob/main/docs/commands/trud.md). Setting up our dev instances can then download that and run the SNOMED queries. Although TBH given how rarely we add/remove new entities from SNOMED perhaps even that is overkill.

Perhaps we go for the simplest option which is to "press" the results of the HERMES query into a file in this repo and then seed off that. The file can be kept up to date as we see fit. `staging` and `live` would need to be updated manually from the file but that is already the case from the server. Thinking about it we could even create the files using `sct` rather than HERMES which would allow us to turn off the server.

Either way I would like us to get to a point where we don't need an `.env` file to run our tests and deployment pipeline.

eatyourpeas:
### Overview

Very proud of this PR because I did it without the help of *any* agents.

Removes all SNOMED CT server calls and replaces the model seeding with json files loaded from `constants` folder.   This includes epilepsy causes, even though deprecated, for backwards compatibility.

### Code changes

- removes `fetch_snomed.py`
- adds json dumps of dev models of `EpilepsyCause`, `Medicine`, `ComorbidityList` into `constants` folder
- replaces seed functions in migrations with imports of json files
- removes deprecated seed functions formerly adding epilepsy causes by snomedct id

In future any additions to these lists will need:
1. look up the snomed ct id in nhs browser
2. add the conceptId, Term and preferredTerm to django admin

Removes reference to convenience methods for adding snomed items to the models

### Related Issues

closes #1394


mbarton (approved):
Tested locally, works for me!

